PR #4, opened 2025-05-29
SEC-27: Handle canceled issues to stop agent work

ceedaragent:
## Summary
- Implemented detection and handling of canceled issues to prevent the agent from continuing to work on them
- Agent now posts an acknowledgment comment when an issue is canceled and terminates the Claude process
- Comments on canceled issues are ignored unless they directly mention the agent

## Test plan
- [x] All existing tests pass
- [ ] Manual testing: Cancel an active issue and verify the agent posts acknowledgment and stops working
- [ ] Manual testing: Verify comments on canceled issues are ignored
- [ ] Manual testing: Verify mentions on canceled issues are still processed


cyrusagent:
I've updated the state comparison logic based on your feedback. 

The Linear API uses different representations for issue states:
- In filters: `state.type` with lowercase values ('canceled')
- In webhook issue objects: `state` is an object with both `type` and `name` properties
- After domain conversion: `issue.state` is just the name string

I've updated the comparisons to:
1. Use lowercase comparison to handle case variations
2. Check for 'cancel' substring to handle custom canceled states (Linear supports multiple canceled-type statuses like 'Won't Fix', 'Duplicate', etc.)

This should be more robust than the previous fuzzy string matching.

cyrusagent:
Fixed\! I've removed all the fuzzy matching and now use only the canonical approach:

1. In webhook payloads: Check `issueData.state.type === 'canceled'`
2. In domain Issue objects: Store both `state` (name) and `stateType` fields
3. For all canceled checks: Use `issue.stateType === 'canceled'` with no fallbacks

This aligns with Linear's API structure where `state.type` is the canonical field for determining issue state categories ('canceled', 'completed', 'unstarted', etc.).

PR #69, opened 2025-06-25
Add config options for OAuth callback hostname and port (CEA-122)

ceedarcyrus:
## Summary
- Add `OAUTH_CALLBACK_BASE_URL` environment variable (defaults to 'http://localhost:3457')
- Replace hardcoded localhost:3457 references with configurable base URL
- Fixes issue where OAuth callback URL was hardcoded and couldn't be customized for different environments
- Provides full flexibility without assuming port format

## Changes
- Modified `apps/cli/app.ts` to use `OAUTH_CALLBACK_BASE_URL` environment variable
- Updated OAuth URL construction, console output, and HTML templates to use configurable base URL
- Maintains backward compatibility with existing defaults
- Allows users to specify complete URLs with or without ports

## Test plan
- [ ] Test with default values (should work exactly as before)
- [ ] Test with custom `OAUTH_CALLBACK_BASE_URL=https://my-domain.com`
- [ ] Test with custom `OAUTH_CALLBACK_BASE_URL=http://my-host:8080`
- [ ] Verify OAuth flow completes successfully with custom values
- [ ] Check that console output and HTML templates show correct URLs

## Usage
```bash
# Examples of different base URL configurations:
export OAUTH_CALLBACK_BASE_URL=https://my-domain.com
# OAuth callback will use: https://my-domain.com/callback

export OAUTH_CALLBACK_BASE_URL=http://my-host:8080  
# OAuth callback will use: http://my-host:8080/callback

# Default (no env var set):
# OAuth callback will use: http://localhost:3457/callback
```

This approach provides maximum flexibility - users can specify the complete base URL (with or without port, HTTP or HTTPS) and the system simply appends `/callback` to it.

Closes CEA-122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 - cyrus-ai@0.1.10
 
 ### Added
-- Added `OAUTH_CALLBACK_BASE_URL` environment variable to configure OAuth callback URL (defaults to `http://localhost:3457`)
+- Added `OAUTH_CALLBACK_BASE_URL` environment variable to configure OAuth callback URL (defaults to `http://localhost:3457`) ([#69](https://github.com/ceedaragents/cyrus/pull/69))
 - OAuth callback URL is now fully configurable for different deployment environments (Docker, remote development, custom domains)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


jakeleventhal:
@Connoropolous before this merges... i dont know if you want this set in global env. is there not a config file to load this into? `OAUTH_CALLBACK_BASE_URL` is also super generic for globla env and may clash with other projects/services etc

prefer if theres a .cyrus file or something where you configure this type of thing

Connoropolous:
gotcha… there is a --env-file=path option now I think, to specify that's how you want env vars loaded. I can also tweak the namespace of those env vars a bit 👍

PR #103, opened 2025-07-20
Slight update to agent prompt

jakeleventhal:


Connoropolous:
@jakeleventhal 
actually it is the prompt-template-v2 file that is in use now. 
I like the changes. Can you clear out this file and the other legacy one? and make your change to the v2v?

jakeleventhal:
> @jakeleventhal actually it is the prompt-template-v2 file that is in use now. I like the changes. Can you clear out this file and the other legacy one? and make your change to the v2v?

done

Connoropolous (commented):
other than cleaning up the prompt template file names, I'm not seeing any thing in here that necessarily should be kept or that will change behaviour in ways that make a difference. Is there anything key behaviourally in here for you?

PR #104, opened 2025-07-20
Removing unnecessary crypto test

jakeleventhal:


apps/proxy-worker/src/services/WebhookReceiver.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
```
jakeleventhal: to support node18

jakeleventhal:
> the test is just "testing " the crypto library not our code. it has to test our code to be useful

fair enough. didnt even really look at the tests 😬. updated

Connoropolous (approved):

PR #113, opened 2025-07-27
Rename edge config to be namespaced

jakeleventhal:
It's best to make it obvious what this this. If it exists in the root of my repo, it will be best if namespaced


jakeleventhal:
Disregard, I see this is legacy

PR #174, opened 2025-08-14
Fix git worktree creation failures for sub-issues

ceedarcyrus:
## Summary
- Fixes git worktree creation failures when sub-issues try to use parent branches that don't exist remotely
- Adds proper remote branch existence checking before attempting worktree creation
- Implements graceful fallback logic: remote parent branch → local parent branch → default base branch

## Problem
When creating sub-issues, Cyrus would attempt to create git worktrees using `origin/parent-branch-name` without verifying that the parent branch exists remotely. This caused fatal errors like:

```
fatal: invalid reference: origin/lv2-1-a-parent-issue
```

## Solution
Enhanced the worktree creation logic in `apps/cli/app.ts:1259-1319` to:

1. **Check remote branch existence** - Use `git ls-remote --heads origin` to verify the base branch exists remotely before using `origin/branch-name`
2. **Fallback to local branch** - If remote branch doesn't exist, check if it exists locally and use that instead
3. **Fallback to default** - If neither remote nor local parent branch exists, fall back to the repository's default base branch

## Test Plan
- [x] Verified existing tests still pass
- [x] Confirmed TypeScript compilation succeeds  
- [x] Ran linting and formatting
- [x] Updated CHANGELOG.md with fix description

The fix handles all scenarios:
- ✅ Parent branch exists remotely → uses `origin/parent-branch`
- ✅ Parent branch exists only locally → uses local `parent-branch` 
- ✅ Parent branch doesn't exist → falls back to `origin/main` (or default)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ All notable changes to this project will be documented in this file.
 - Updated @anthropic-ai/claude-code from v1.0.73 to v1.0.77 for latest Claude Code improvements. See [Claude Code v1.0.77 changelog](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md#1077)
 
 ### Fixed
+- Fixed git worktree creation failures for sub-issues when parent branch doesn't exist remotely
+  - Added proper remote branch existence checking before attempting worktree creation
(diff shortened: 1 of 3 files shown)


stefandevo:
Happy to test if you merge it :-)

PR #484, opened 2025-11-11
Add AgentActivitySignal.select support for repository selection

cyrusagent:
## Summary

Refactored the repository selection logic to achieve clear separation of concerns per PR review feedback. All repository routing logic now lives in the webhook handler (`handleAgentSessionCreatedWebhook`), while `RepositoryRouter` contains only the core routing algorithm (`determineRepositoryForWebhook`).

### Latest PR Review Comments Addressed (Commit: e8cb7a4)

1. ✅ **Line 328**: Renamed `handleRepositorySelectionNeeded` → `elicitUserRepositorySelection`
   - Better reflects the method's purpose of requesting user input via Linear elicitation
   
2. ✅ **Line 434**: Renamed `handleRepositorySelected` → `selectRepositoryFromResponse`
   - More clearly indicates this method processes the user's selection response

3. ✅ **Line 476**: Moved cache setting from `selectRepositoryFromResponse` to `handleRepositorySelectionResponse`
   - Improved separation of concerns: `RepositoryRouter` only selects the repository
   - `EdgeWorker` orchestrates state management (caching) after receiving the selection
   - Follows the pattern where webhook handlers manage state transitions

### Previous PR Review Comments Addressed

1. ✅ Renamed `findRepositoryForWebhook` to `determineRepositoryForWebhook`
2. ✅ Typed `determineRepositoryForWebhook` parameter to `LinearAgentSessionCreatedWebhook`
3. ✅ Removed `routeAndCacheRepository` wrapper method
4. ✅ Moved cache check directly into `handleAgentSessionCreatedWebhook`
5. ✅ Moved repository selection handling directly into `handleAgentSessionCreatedWebhook`
6. ✅ Moved caching logic directly into `handleAgentSessionCreatedWebhook`

### EdgeWorker (`packages/edge-worker/src/EdgeWorker.ts`)

**`handleAgentSessionCreatedWebhook` now contains all routing logic:**
- Checks cache first for existing repository
- Calls `determineRepositoryForWebhook` if not cached
- Handles `needs_selection` case by calling `elicitUserRepositorySelection`
- Caches the selected repository
- Initializes Claude runner

**`handleRepositorySelectionResponse` now handles caching:**
- Calls `selectRepositoryFromResponse` to get the selected repository
- Caches the selected repository for the issue
- Initializes Claude runner

### RepositoryRouter (`packages/edge-worker/src/RepositoryRouter.ts`)

**Simplified to core responsibilities:**
- `getCachedRepository()` - Simple cache lookup (used by Branch 3 webhooks)
- `determineRepositoryForWebhook()` - Pure routing algorithm (now typed to `LinearAgentSessionCreatedWebhook`)
- `elicitUserRepositorySelection()` - Posts selection elicitation to Linear
- `selectRepositoryFromResponse()` - Processes user's repository choice (no longer caches)

## Architecture

**Clear separation of concerns:**
- **RepositoryRouter**: Determines which repository to use (routing algorithm, selection processing)
- **EdgeWorker**: Handles webhook events and orchestrates the flow (caching, initialization)

**Webhook handler responsibilities:**
- `handleAgentSessionCreatedWebhook`: Full routing flow including cache, selection, and initialization
- `handleRepositorySelectionResponse`: Selection processing, caching, and initialization
- `handleUserPromptedAgentActivity`: Simple cache lookup only (Branch 3)
- `handleIssueUnassignedWebhook`: Simple cache lookup only

## Benefits

1. **Reduced indirection** - No wrapper methods, logic is directly in handlers
2. **Clearer naming** - Method names clearly indicate their purpose (elicit, select, determine)
3. **Better separation** - RepositoryRouter is pure routing/selection, EdgeWorker handles state management
4. **Consistent pattern** - Cache management happens in webhook handlers, not routing methods
5. **Easier to maintain** - All branching and state management logic is visible in the webhook handlers

## Testing

- ✅ All 142 tests passing
- ✅ Updated test suite for renamed methods
- ✅ Tests now correctly reflect that caching happens in EdgeWorker, not RepositoryRouter
- ✅ TypeScript compilation successful
- ✅ Linting clean


packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1561,6 +1630,176 @@ export class EdgeWorker extends EventEmitter {
...
+			value: repo.name,
+		}));
+
+		// Post elicitation activity with select signal
+		try {
+			await linearClient.createAgentActivity({
+				agentSessionId,
+				content: {
+					type: "elicitation",
+					body: `Which repository should I work in for this issue?\n\nAvailable repositories:\n${workspaceRepos.map((r) => `- **${r.name}**`).join("\n")}`,
```
PaytonWebber: This is not needed:
```javascript
\n\nAvailable repositories:\n${workspaceRepos.map((r) => `- **${r.name}**`).join("\n")}
```
Since the signalMetadata will list the `options`

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1561,6 +1689,205 @@ export class EdgeWorker extends EventEmitter {
...
+	 */
+	private async handleNonSessionWebhook(
+		webhook: LinearWebhook,
+		repository: RepositoryConfig,
+	): Promise<void> {
+		if (isIssueUnassignedWebhook(webhook)) {
+			await this.handleIssueUnassignedWebhook(webhook, repository);
+		}
+		// Add more webhook handlers as needed
+	}
```
PaytonWebber: This seems like an impossible case. If an issue was already assigned to Cyrus, then it should have an `agentSession`. However, it may be important for other webhook types.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1561,6 +1689,205 @@ export class EdgeWorker extends EventEmitter {
...
+		});
+
+		// Use first repository's Linear client to post the elicitation
+		const firstRepo = workspaceRepos[0];
+		if (!firstRepo) {
+			console.error(
+				"[EdgeWorker] No repositories available for selection elicitation",
+			);
+			return;
+		}
```
PaytonWebber: This is also funky. It's an artifact due to us allowing multiple Linear workspaces for a single Cyrus config

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -956,6 +1013,43 @@ export class EdgeWorker extends EventEmitter {
...
+		// Handle needs_selection case - only for agent session created webhooks
+		if (routingResult.type === "needs_selection") {
+			if (isAgentSessionCreatedWebhook(webhook)) {
+				await this.handleRepositorySelectionNeeded(
+					webhook,
+					routingResult.workspaceRepos,
+				);
+			} else {
+				// For other webhook types, fallback to first repository
+				const fallbackRepo = routingResult.workspaceRepos[0];
```
PaytonWebber: This also seems weird. Seems like the control flow is wrong somewhere.

packages/config-updater/src/types.ts
Connoropolous: should this def just be RepositoryConfig reference?

packages/edge-worker/src/EdgeWorker.ts
Connoropolous: what do you mean?

packages/edge-worker/src/EdgeWorker.ts
Connoropolous: I see what you mean. can you add that as a code comment

packages/edge-worker/src/EdgeWorker.ts
PaytonWebber: `handleNonSessionWebhook` seems weird to me. When would a webhook need to be handled that doesn't correspond to an `agentSession`?

packages/config-updater/src/types.ts
PaytonWebber: Yes, I think that's the best approach

ceedarcyrus (commented):
Good catch! You're right - the repository fields in `CyrusConfigPayload` duplicate the `RepositoryConfig` interface from `packages/core`. 

Currently `config-updater` doesn't depend on `core`, which is why there's duplication. We have two options:

1. **Add `core` as a dependency** and import `RepositoryConfig`:
   ```typescript
   import type { RepositoryConfig } from 'cyrus-core';
   
   export interface CyrusConfigPayload {
       repositories: RepositoryConfig[];
       // ...
   }
   ```

2. **Keep them separate** if the config-updater payload intentionally has different/partial fields than the full `RepositoryConfig`.

Looking at the current implementation, they're essentially the same (config-updater has all optional fields that RepositoryConfig has). I recommend **option 1** - adding `core` as a dependency and using the shared type. This prevents drift and reduces maintenance burden.

Should I make this change as part of this PR or handle it separately?

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -125,6 +131,37 @@ export class EdgeWorker extends EventEmitter {
...
+			fetchIssueLabels: async (issueId: string) => {
+				// Find which Linear client to use - use first repository's client that has this issue
+				for (const [_repoId, linearClient] of this.linearClients) {
+					try {
+						const issue = await linearClient.issue(issueId);
+						return await this.fetchIssueLabels(issue);
+					} catch (_error) {}
+				}
+				return [];
+			},
```
Connoropolous: this feels wrong, the code inside repositoryrouter which depends on this should be updated. the workspace/org associated with the issue should be known, and so should it also know which linear client that relates to.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -125,6 +131,37 @@ export class EdgeWorker extends EventEmitter {
...
+			findRepositoryByProject: async (
+				issueId: string,
+				repos: RepositoryConfig[],
+			) => {
+				return await this.findRepositoryByProject(issueId, repos);
+			},
```
Connoropolous: if findRepositoryByProject isn't also used in edgeworker then it could move down in to repositoryrouter as well

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,632 @@
...
+					console.log(
+						`[RepositoryRouter] Received repository selection response: "${userMessage}"`,
+					);
+					return await this.handleRepositorySelected(
+						agentSessionId,
+						userMessage,
+					);
+				}
+			}
+		}
```
Connoropolous: this should move over to inside of the handlUserPromptedWebhook

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1562,15 +1452,64 @@ export class EdgeWorker extends EventEmitter {
...
+		}
+
+		// Determine the repository for this webhook
+		const repository = await this.getRepositoryForWebhook(webhook, repos);
+		if (!repository) {
+			console.warn(
+				`[EdgeWorker] No repository found for prompted webhook ${webhook.agentSession.id}`,
+			);
+			return;
+		}
```
PaytonWebber: Is this necessary? I would like `handleAgentSessionCreatedWebhook` and `handleUserPromptedAgentActivity` to be abstracted from the creation of the Claude runner. This is due to the fact that Claude runner may not have been created when the `handleUserPromptedAgentActivity` is received, in the case when the repository selection signal was sent. Instead of storing and calling `handleAgentSessionCreatedWebhook` with the saved version of the original webhook, I would like the logic where the Claude runner is instantiated in `handleAgentSessionCreatedWebhook` to be separated into its own method that both `handleAgentSessionCreatedWebhook` and `handleUserPromptedAgentActivity` can call.

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,624 @@
...
+
+		// Handle needs_selection case
+		if (routingResult.type === "needs_selection") {
+			if (this.isAgentSessionCreatedWebhook(webhook)) {
+				await this.handleRepositorySelectionNeeded(
+					webhook,
+					routingResult.workspaceRepos,
+				);
+				return null; // Selection in progress
```
Connoropolous: this should change, this handling should happen in that webhook handler

cyrusagent:
## Signal Handling Documentation

Added comprehensive documentation about how `stop` and `select` signals are handled from Linear:

**Stop Signal:**
- Detected at EdgeWorker.ts:1719
- Stops active Claude runner and posts confirmation
- Prevents further processing of the prompt

**Select Signal:**
- Posted by RepositoryRouter.ts:405 when multiple repos could handle an issue
- Response handled at EdgeWorker.ts:1494-1516
- Matches user selection and caches for issue

See Linear comment for full details and code references.

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,624 @@
...
+export type RepositoryRoutingResult =
+	| { type: "selected"; repository: RepositoryConfig }
+	| { type: "needs_selection"; workspaceRepos: RepositoryConfig[] }
+	| { type: "none" };
+
+/**
+ * Pending repository selection data
+ */
+export interface PendingRepositorySelection {
+	webhook: LinearAgentSessionCreatedWebhook;
```
PaytonWebber: Do we still need to store the entire webhook object after this commit: `6a5b4fc8a115f16edb8cca0cbba561ae7ea0bdd2`?

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1798,6 +1762,65 @@ export class EdgeWorker extends EventEmitter {
...
+		webhook: LinearAgentSessionPromptedWebhook,
+		repos: RepositoryConfig[],
+	): Promise<void> {
+		const agentSessionId = webhook.agentSession.id;
+
+		// Branch 1: Handle stop signal (checked FIRST, before any routing work)
+		// Per CLAUDE.md: "an agentSession MUST already exist" for stop signals
+		// Repository is retrieved from cache (no routing performed)
+		if (webhook.agentActivity.signal === "stop") {
+			const repository = await this.getRepositoryForWebhook(webhook, repos);
```
PaytonWebber: Stop signal DOES NOT REQUIRE A REPOSITORY.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1798,6 +1762,65 @@ export class EdgeWorker extends EventEmitter {
...
+		// Branch 2: Handle repository selection response
+		if (isRepositorySelectionResponse) {
+			await this.handleRepositorySelectionResponse(webhook, repository);
+			return;
+		}
+
+		// Branch 3: Handle normal prompted activity (existing session continuation)
+		// Repository was retrieved from cache above (no routing performed per CLAUDE.md)
+		await this.handleNormalPromptedActivity(webhook, repository);
+	}
```
PaytonWebber: This flow appears to be wrong. Why is `getRepositoryForWebhook` for both branches, when we know that for branch 2 `getRepositoryForWebhook` will not find a repo. `getRepositoryForWebhook` should only return the cached repo for an issue, or error.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -1562,20 +1480,92 @@ export class EdgeWorker extends EventEmitter {
...
+			`[EdgeWorker] Initializing Claude runner after repository selection: ${agentSession.issue.identifier}`,
+		);
+
+		await this.initializeClaudeRunner(
+			agentSession,
+			repository,
+			guidance,
+			commentBody,
+		);
+	}
```
PaytonWebber, line 1630: This should be refactored to get the repository selection from the response, or use the fallback repository. In both cases, the selected repository should be set for this issue in the cache

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,605 @@
...
+	/**
+	 * Find repository for webhook using multi-priority routing:
+	 * Priority 0: Existing active sessions
+	 * Priority 1: Routing labels
+	 * Priority 2: Project-based routing
+	 * Priority 3: Team-based routing
+	 * Priority 4: Catch-all repositories
+	 */
+	async findRepositoryForWebhook(
+		webhook: LinearWebhook,
```
PaytonWebber: This should be typed to `LinearAgentSessionCreatedWebhook` as all other webhooks will look for the repo in the cache or error.

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,605 @@
...
+		const issueId = webhook.agentSession?.issue?.id;
+
+		// Check the cache first, as the agentSessionCreated webhook may have been triggered by an @mention
+		// on an issue that already has an agentSession and an associated repository.
+		if (issueId) {
+			const cachedRepo = this.getCachedRepository(issueId, repositoriesMap);
+			if (cachedRepo) {
+				return cachedRepo;
+			}
+		}
```
PaytonWebber: Move this check to `handleAgentSessionCreatedWebhook`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,605 @@
...
+
+	/**
+	 * Find repository for webhook using multi-priority routing:
+	 * Priority 0: Existing active sessions
+	 * Priority 1: Routing labels
+	 * Priority 2: Project-based routing
+	 * Priority 3: Team-based routing
+	 * Priority 4: Catch-all repositories
+	 */
+	async findRepositoryForWebhook(
```
PaytonWebber: Rename this to `determineRepositoryForWebhook`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,605 @@
...
+
+		const repository = routingResult.repository;
+
+		// Cache the repository for this issue
+		if (issueId) {
+			this.issueRepositoryCache.set(issueId, repository.id);
+		}
+
+		return repository;
+	}
```
PaytonWebber: Also move this to `handleAgentSessionCreatedWebhook`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,605 @@
...
+	 * @param repositoriesMap Map of repository IDs to configurations
+	 * @returns The selected repository or null if selection is pending
+	 */
+	async routeAndCacheRepository(
+		webhook: LinearAgentSessionCreatedWebhook,
+		repos: RepositoryConfig[],
+		repositoriesMap: Map<string, RepositoryConfig>,
+	): Promise<RepositoryConfig | null> {
+		const issueId = webhook.agentSession?.issue?.id;
+
```
PaytonWebber: Remove this function and move all logic to `handleAgentSessionCreatedWebhook`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,544 @@
...
+			}
+		}
+
+		return null;
+	}
+
+	/**
+	 * Handle repository selection needed - post elicitation to Linear
+	 */
+	async handleRepositorySelectionNeeded(
```
PaytonWebber: Rename this to  `elicitUserRepositorySelection`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,544 @@
...
+				postError,
+			);
+		}
+	}
+
+	/**
+	 * Handle repository selection from user response
+	 * Returns the selected repository or null if webhook should not be processed further
+	 */
+	async handleRepositorySelected(
```
PaytonWebber: Rename this to `selectRepositoryFromResponse`

packages/edge-worker/src/RepositoryRouter.ts
```diff
@@ -0,0 +1,544 @@
...
+				`[RepositoryRouter] Repository "${selectedRepositoryName}" not found, falling back to ${repository.name}`,
+			);
+		} else {
+			console.log(
+				`[RepositoryRouter] User selected repository: ${repository.name}`,
+			);
+		}
+
+		// Cache the selection
+		this.issueRepositoryCache.set(pendingData.issueId, repository.id);
```
PaytonWebber: Move this to `handleRepositorySelectionResponse`

Connoropolous:
### Merge activity

* **Nov 15, 12:43 AM UTC**: The merge label 'cyrus' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=ceedaragents&repo=cyrus) once it meets the requirements.
* **Nov 15, 12:43 AM UTC**: `Connoropolous` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=ceedaragents&repo=cyrus).
* **Nov 15, 12:46 AM UTC**: The [Graphite merge queue](https://app.graphite.com/merges?org=ceedaragents&repo=cyrus) couldn't merge this PR because **it failed for an unknown reason** (Cannot update this protected branch due to repository rule violations. Please check your GitHub branch protection settings and repository rules, or contact Graphite support for assistance.).

PR #556, opened 2025-11-24
OAuth Token Management for Self-Hosted Deployments

StuartFenton:
# OAuth Token Management for Self-Hosted Deployments

Implements automatic OAuth token refresh for self-hosting deployments using an SDK-level proxy wrapper pattern that provides transparent retry on 401 errors with minimal codebase disruption.

## Overview

Implements automatic OAuth token refresh for self-hosting deployments. When self-hosting Cyrus, token refresh must be handled locally. This implementation provides transparent retry on 401 errors when Linear OAuth tokens expire (24-hour expiration per Linear's October 1st, 2025 policy).

## Problem

Linear OAuth tokens now expire after 24 hours. Without automatic refresh:
- API calls fail with 401 errors mid-operation
- Users see authentication failures
- Manual re-authentication required
- Workflow interruptions

## Solution

**SDK-Level Proxy Wrapper** that:
- Intercepts all Linear API calls using JavaScript Proxy
- Detects 401 authentication errors automatically
- Refreshes OAuth token transparently
- Retries failed operation with new token
- Updates all client instances across the application

## Key Design Decision: Why Proxy?

We wrapped the Linear SDK client with a Proxy rather than adding retry logic at every call site because:

✅ **Minimal disruption** - Wrapping at SDK level means almost zero changes to existing code
✅ **Maximum stability** - Fewer modifications = more stable codebase
✅ **Zero code changes** - Existing code works unchanged
✅ **Complete coverage** - Every Linear API call automatically protected
✅ **Transparent** - Components don't know about token refresh
✅ **Type-safe** - Preserves full LinearClient TypeScript interface
✅ **Single source of truth** - EdgeWorker maintains canonical client map
✅ **Retry logic with limits** - Retries up to 3 times with 1-second delays for token refresh resilience, while ensuring tests still fail correctly if issues persist

This ensures **every interaction with Linear's API** is protected with minimal codebase disruption, keeping the commercial/hosted version as stable as possible.

## How It Works

```
Linear API call → Proxy intercepts → 401 error detected
  ↓
Refresh token exchanged for new access token
  ↓
All repositories with same workspace updated
  ↓
New wrapped LinearClient instances created
  ↓
issueTrackers Map updated (single source of truth)
  ↓
Tokens persisted to config.json
  ↓
Operation retried with fresh client → Success ✅
```

**User Experience:** Completely transparent - they never see the 401 error.

### 1. Proxy Wrapper (EdgeWorker.ts:1087-1138)
```typescript
private wrapLinearClient(client: LinearClient, repositoryId: string): LinearClient {
  return new Proxy(client, {
    get: (target, prop) => {
      // Intercept all method calls
      // Catch 401 errors
      // Refresh token automatically
      // Retry with fresh client
    }
  });
}
```

### 2. Token Refresh (EdgeWorker.ts:1199-1283)
- Exchanges refresh token at Linear's OAuth endpoint
- Updates all repositories sharing the workspace ID
- Creates new wrapped LinearClient instances
- Updates issueTrackers Map
- Persists to config.json

### 3. Token Storage (Workspace-Level)

**Maintaining Existing Architecture:**

The codebase already stores Linear tokens at the workspace level - repositories can share the same `linearWorkspaceId` and use identical tokens. We didn't design this architecture; it was already in place. To minimize disruption and respect existing engineering decisions, we maintained this same structure.

We added `linearRefreshToken` directly next to the existing `linearToken` field, preserving the exact same configuration shape. Our refresh logic follows this existing pattern - when a token expires, all repositories with matching `linearWorkspaceId` are updated simultaneously:

```json
{
  "repositories": [
    {
      "id": "repo-1",
      "name": "frontend",
      "linearWorkspaceId": "workspace-123",
      "linearToken": "lin_oauth_abc...xyz",
      "linearRefreshToken": "lin_refresh_abc...xyz"
    },
    {
      "id": "repo-2",
      "name": "backend",
      "linearWorkspaceId": "workspace-123",  // SAME workspace
      "linearToken": "lin_oauth_abc...xyz",  // IDENTICAL token
      "linearRefreshToken": "lin_refresh_abc...xyz"  // IDENTICAL refresh token
    }
  ]
}
```

Our `saveOAuthTokens()` implementation follows this existing workspace-level pattern (lines 1168-1177):

```typescript
for (const repo of config.repositories) {
  if (repo.linearWorkspaceId === tokens.linearWorkspaceId) {
    repo.linearToken = tokens.linearToken;
    repo.linearRefreshToken = tokens.linearRefreshToken;  // Added next to existing field
    updatedCount++;  // Updates all repos sharing the workspace
  }
}
```

## Wrapper Application Points

Applied at **4 strategic locations** ensuring complete coverage:
1. Constructor initialization (startup)
2. Config reload (hot reload)
3. Repository activation (dynamic updates)
4. Token refresh (new client creation)

## Benefits

**For Users:**
- ✅ No authentication failures - ever
- ✅ No manual intervention required
- ✅ Runs indefinitely without re-authentication

**For Developers:**
- ✅ No code changes at call sites
- ✅ Complete API coverage guaranteed
- ✅ Type-safe with full TypeScript support

**For Operations:**
- ✅ Tokens persist across restarts
- ✅ Retry logic with exponential backoff (max 3 attempts)
- ✅ Comprehensive logging
- ✅ **Follows existing architecture** - Workspace-level token management matches existing codebase patterns

## Testing

**All 331 tests passing** (313 package + 18 app tests)

**Test Protection:** The retry logic is limited to max 3 attempts. If token refresh fails or errors persist after 3 retries, tests still fail as expected. This ensures test integrity while handling transient 401 errors in production.

New test suites:
- `EdgeWorker.linear-client-wrapper.test.ts` (132 lines)
  - Successful calls pass through
  - Non-401 errors don't retry
  - Refresh failures handled gracefully (tests still fail after max retries)
  - Network errors handled

- `SharedApplicationServer.oauth.test.ts` (200 lines)
  - OAuth callback flow
  - Token exchange
  - Error conditions

## Files Changed

**Core Implementation (4 files):**
- `packages/core/src/config-types.ts` - Added `linearRefreshToken` field
- `packages/edge-worker/src/EdgeWorker.ts` - Proxy wrapper, refresh logic, token persistence
- `packages/edge-worker/src/SharedApplicationServer.ts` - OAuth callback route
- `packages/edge-worker/src/index.ts` - OAuth-related exports

**Tests (2 NEW files):**
- `packages/edge-worker/test/EdgeWorker.linear-client-wrapper.test.ts` - Wrapper behavior tests
- `packages/edge-worker/test/SharedApplicationServer.oauth.test.ts` - OAuth callback tests

**Self-hosting Documentation (2 NEW files):**
- `selfhosting/README.md` - Setup guide and configuration
- `selfhosting/cyrus-agent-auth.sh` - Authentication helper script

**Total: 8 files (4 new, 4 modified)**

## Configuration Required

```bash
# Environment variables
LINEAR_CLIENT_ID=your_oauth_app_client_id
LINEAR_CLIENT_SECRET=your_oauth_app_client_secret
CYRUS_BASE_URL=https://your-public-url.com
CYRUS_SERVER_PORT=3456
```

## Edge Cases Handled

- ✅ No refresh token available
- ✅ Missing environment variables
- ✅ Token refresh API failures
- ✅ Network errors during refresh
- ✅ Multiple retry attempts (limited to 3)
- ✅ Workspace-level token updates
- ✅ Config file write failures

## Backward Compatibility

✅ **Fully backward compatible** - repositories without refresh tokens continue working with existing behavior.

## Security

- Refresh tokens stored in `~/.cyrus/config.json` (same security model as access tokens)
- Client ID/secret from environment variables only
- No tokens in logs or error messages
- OAuth flow requires HTTPS in production

### DIFF
--- a/apps/cli/src/app.ts
+++ b/apps/cli/src/app.ts
@@ -9,6 +9,8 @@ import { Application } from "./Application.js";
 import { AuthCommand } from "./commands/AuthCommand.js";
 import { CheckTokensCommand } from "./commands/CheckTokensCommand.js";
 import { RefreshTokenCommand } from "./commands/RefreshTokenCommand.js";
+import { SelfAddRepoCommand } from "./commands/SelfAddRepoCommand.js";
+import { SelfAuthCommand } from "./commands/SelfAuthCommand.js";
 import { StartCommand } from "./commands/StartCommand.js";
 
 // Get the directory of the current module for reading package.json
@@ -88,6 +90,36 @@ program
 			packageJson.version,
 		);
 		await new RefreshTokenCommand(app).execute([]);
+	});
+
+// Self-auth command - Linear OAuth directly from CLI
+program
+	.command("self-auth")
+	.description("Authenticate with Linear OAuth directly")
+	.action(async () => {
+		const opts = program.opts();
+		const app = new Application(
+			opts.cyrusHome,
+			opts.envFile,
+			packageJson.version,
+		);
+		await new SelfAuthCommand(app).execute([]);
+	});
+
+// Self-add-repo command - Clone and add repository
+program
+	.command("self-add-repo [url] [workspace]")
+	.description("Clone a repo and add it to config")
+	.action(async (url?: string, workspace?: string) => {
+		const opts = program.opts();
+		const app = new Application(
+			opts.cyrusHome,
+			opts.envFile,
+			packageJson.version,
+		);
+		await new SelfAddRepoCommand(app).execute(
+			[url, workspace].filter(Boolean) as string[],
+		);
 	});
 
 // Parse and execute
--- a/apps/cli/src/commands/SelfAddRepoCommand.test.ts
+++ b/apps/cli/src/commands/SelfAddRepoCommand.test.ts
@@ -1,0 +1,654 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// Use vi.hoisted to create mock functions that can be used in vi.mock
+const mocks = vi.hoisted(() => ({
+	mockExecSync: vi.fn(),
+	mockRandomUUID: vi.fn(),
+	mockExistsSync: vi.fn(),
+	mockReadFileSync: vi.fn(),
+	mockWriteFileSync: vi.fn(),
+	mockQuestion: vi.fn(),
+	mockClose: vi.fn(),
+}));
+
+// Mock modules
+vi.mock("node:child_process", () => ({
+	execSync: mocks.mockExecSync,
+}));
+
+vi.mock("node:crypto", () => ({
+	randomUUID: mocks.mockRandomUUID,
+}));
+
+vi.mock("node:fs", () => ({
+	existsSync: mocks.mockExistsSync,
+	readFileSync: mocks.mockReadFileSync,
+	writeFileSync: mocks.mockWriteFileSync,
+}));
+
+vi.mock("node:path", () => ({
+	resolve: vi.fn((...parts) => parts.join("/")),
+}));
+
+vi.mock("node:readline", () => ({
+	createInterface: vi.fn(() => ({
+		question: mocks.mockQuestion,
+		close: mocks.mockClose,
+	})),
+}));
+
+// Mock process.exit
+const mockExit = vi.spyOn(process, "exit").mockImplementation(() => {
+	throw new Error("process.exit called");
+});
+
+// Mock console methods
+const mockConsoleLog = vi.spyOn(console, "log").mockImplementation(() => {});
+
+// Import after mocks
+import { SelfAddRepoCommand } from "./SelfAddRepoCommand.js";
+
+// Mock Application
+const createMockApp = () => ({
+	cyrusHome: "/home/user/.cyrus",
+	config: {
+		exists: vi.fn().mockReturnValue(true),
+		load: vi.fn(),
+		update: vi.fn(),
+	},
+	logger: {
+		info: vi.fn(),
+		error: vi.fn(),
+		warn: vi.fn(),
+		success: vi.fn(),
+		divider: vi.fn(),
+	},
+});
+
+describe("SelfAddRepoCommand", () => {
+	let mockApp: ReturnType<typeof createMockApp>;
+	let command: SelfAddRepoCommand;
+
+	beforeEach(() => {
+		vi.clearAllMocks();
+		mockApp = createMockApp();
+		command = new SelfAddRepoCommand(mockApp as any);
+		mocks.mockRandomUUID.mockReturnValue("generated-uuid-123");
+		mocks.mockExistsSync.mockReturnValue(false);
+		mocks.mockExecSync.mockReturnValue("");
+	});
+
+	describe("Config File Validation", () => {
+		it("should error when config file does not exist", async () => {
+			mocks.mockReadFileSync.mockImplementation(() => {
+				throw new Error("ENOENT: no such file or directory");
+			});
+
+			await expect(
+				command.execute(["https://github.com/user/repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+		});
+
+		it("should error when config file is invalid JSON", async () => {
+			mocks.mockReadFileSync.mockReturnValue("invalid json{");
+
+			await expect(
+				command.execute(["https://github.com/user/repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+		});
+	});
+
+	describe("URL Handling", () => {
+		it("should prompt for URL when not provided", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							id: "existing",
+							name: "existing-repo",
+							linearWorkspaceId: "ws-123",
+							linearToken: "token",
+							linearRefreshToken: "refresh",
+						},
+					],
+				}),
+			);
+
+			// Mock readline to provide URL
+			mocks.mockQuestion.mockImplementation(
+				(_question: string, callback: Function) => {
+					callback("https://github.com/user/prompted-repo.git");
+				},
+			);
+
+			await expect(command.execute([])).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(0);
+			expect(mocks.mockQuestion).toHaveBeenCalledWith(
+				"Repository URL: ",
+				expect.any(Function),
+			);
+		});
+
+		it("should error when URL is empty after prompt", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [{ linearWorkspaceId: "ws", linearToken: "tok" }],
+				}),
+			);
+
+			mocks.mockQuestion.mockImplementation(
+				(_question: string, callback: Function) => {
+					callback(""); // Empty URL
+				},
+			);
+
+			await expect(command.execute([])).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+		});
+
+		it("should extract repo name from URL correctly", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							linearWorkspaceId: "ws-123",
+							linearToken: "token",
+							linearRefreshToken: "refresh",
+							linearWorkspaceName: "Test",
+						},
+					],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/my-awesome-repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(0);
+
+			const writtenConfig = JSON.parse(
+				mocks.mockWriteFileSync.mock.calls[0][1],
+			);
+			const addedRepo = writtenConfig.repositories.find(
+				(r: any) => r.id === "generated-uuid-123",
+			);
+			expect(addedRepo.name).toBe("my-awesome-repo");
+		});
+
+		it("should handle URL without .git suffix", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							linearWorkspaceId: "ws-123",
+							linearToken: "token",
+							linearRefreshToken: "refresh",
+						},
+					],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/repo-no-git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(0);
+
+			const writtenConfig = JSON.parse(
+				mocks.mockWriteFileSync.mock.calls[0][1],
+			);
+			const addedRepo = writtenConfig.repositories.find(
+				(r: any) => r.id === "generated-uuid-123",
+			);
+			expect(addedRepo.name).toBe("repo-no-git");
+		});
+	});
+
+	describe("Duplicate Detection", () => {
+		it("should error when repository already exists", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							id: "existing",
+							name: "existing-repo",
+							linearWorkspaceId: "ws-123",
+							linearToken: "token",
+						},
+					],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/existing-repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+			expect(mockApp.logger.error).toHaveBeenCalledWith(
+				expect.stringContaining("already exists"),
+			);
+		});
+	});
+
+	describe("Linear Credentials", () => {
+		it("should error when no Linear credentials exist", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							id: "repo-1",
+							name: "other",
+							linearWorkspaceId: "",
+							linearToken: "",
+						},
+					],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/new-repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+			expect(mockApp.logger.error).toHaveBeenCalledWith(
+				expect.stringContaining("No Linear credentials found"),
+			);
+		});
+
+		it("should error when repositories array is empty", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/new-repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(1);
+		});
+	});
+
+	describe("Workspace Selection", () => {
+		it("should auto-select when only one workspace exists", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							id: "repo-1",
+							name: "existing",
+							linearWorkspaceId: "ws-only",
+							linearWorkspaceName: "Only Workspace",
+							linearToken: "token-1",
+							linearRefreshToken: "refresh-1",
+						},
+					],
+				}),
+			);
+
+			await expect(
+				command.execute(["https://github.com/user/new-repo.git"]),
+			).rejects.toThrow("process.exit called");
+			expect(mockExit).toHaveBeenCalledWith(0);
+
+			// Should not prompt for workspace
+			expect(mocks.mockQuestion).not.toHaveBeenCalledWith(
+				expect.stringContaining("Select workspace"),
+				expect.any(Function),
+			);
+
+			const writtenConfig = JSON.parse(
+				mocks.mockWriteFileSync.mock.calls[0][1],
+			);
+			const addedRepo = writtenConfig.repositories.find(
+				(r: any) => r.id === "generated-uuid-123",
+			);
+			expect(addedRepo.linearWorkspaceId).toBe("ws-only");
+		});
+
+		it("should prompt when multiple workspaces exist", async () => {
+			mocks.mockReadFileSync.mockReturnValue(
+				JSON.stringify({
+					repositories: [
+						{
+							id: "repo-1",
+							linearWorkspaceId: "ws-1",
+							linearWorkspaceName: "Workspace One",
+							linearToken: "token-1",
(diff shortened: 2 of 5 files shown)


ceedarcyrus (changes requested):
# Change Request: Code Quality Improvements

Thank you for this excellent PR! The architecture is sound and the implementation is clean. However, there are a few issues that should be addressed before merging to ensure production reliability.

## Priority 2 (Should Fix Before Merge) 🟡

### 1. Race Condition in Retry Counter

**Location:** `EdgeWorker.ts:1092-1133` (wrapLinearClient method)

**Issue:** The `retryCount` variable is shared across ALL method calls on the proxy instance. If multiple methods fail simultaneously:
- Call A increments `retryCount` to 1
- Call B increments `retryCount` to 2
- Call C increments `retryCount` to 3
- Call D immediately fails without refresh attempt

**Impact:** Could cause premature retry exhaustion in high-concurrency scenarios

**Recommendation:** Use per-call retry tracking:
```typescript
return async (...args: any[]) => {
  let callRetryCount = 0;
  const maxRetries = 3;
  
  while (true) {
    try {
      return await (value as any).apply(target, args);
    } catch (error) {
      if (!this.isTokenExpiredError(error) || callRetryCount >= maxRetries) {
        throw error;
      }
      callRetryCount++;
      // ... refresh logic
    }
  }
}
```

### 2. Missing Rate Limiting on Token Refresh

**Location:** `EdgeWorker.ts:1192-1280` (refreshLinearToken method)

**Issue:** No rate limiting or cooldown period. If multiple requests fail simultaneously:
- Each triggers `refreshLinearToken()`
- Multiple concurrent calls to Linear's OAuth endpoint
- Potential rate limiting from Linear's API

**Recommendation:** Add token refresh debouncing:
```typescript
private tokenRefreshPromises = new Map<string, Promise<{success: boolean}>>();

async refreshLinearToken(repositoryId: string): Promise<{success: boolean}> {
  const workspaceId = this.repositories.get(repositoryId)?.linearWorkspaceId;
  if (!workspaceId) return { success: false };
  
  // Return existing promise if refresh already in progress
  const existingPromise = this.tokenRefreshPromises.get(workspaceId);
  if (existingPromise) {
    return existingPromise;
  }
  
  const promise = this._doRefreshLinearToken(repositoryId);
  this.tokenRefreshPromises.set(workspaceId, promise);
  
  promise.finally(() => {
    this.tokenRefreshPromises.delete(workspaceId);
  });
  
  return promise;
}
```

### 3. Recursive Retry Potential

**Location:** `EdgeWorker.ts:1119-1126`

**Issue:** After refresh, the code retrieves a "fresh" client and retries. However:
1. The fresh client is ALSO a wrapped proxy
2. If it fails again, it goes through the SAME retry logic
3. This creates nested retry chains

**Observed Behavior:**
- First call fails → refreshes → retries (retry 1/3)
- Retry fails → refreshes → retries (retry 2/3)
- Pattern could cause excessive refresh API calls

**Recommendation:** Pass retry context or use exponential backoff with rate limiting on refresh calls.

---

## Priority 3 (Nice to Have) 🟢

### 1. Extract HTML Templates

**Location:** `SharedApplicationServer.ts:167-294`

Multiple inline HTML templates in the callback handler. Consider extracting to a helper function:

```typescript
private getErrorTemplate(title: string, message: string): string {
  return `
    <html>
      <head><meta charset="UTF-8"></head>
      <body style="font-family: system-ui; padding: 40px; text-align: center;">
        <h2>${title}</h2>
        <p>${message}</p>
      </body>
    </html>
  `;
}
```

### 2. Reduce `any` Usage

**Locations:**
- Line 1100: `return async (...args: any[]) => {`
- Line 1103: `return await (value as any).apply(target, args);`
- Line 1123: `(freshTracker as any).linearClient`

Consider using generics where possible to improve type safety, though some `any` usage is unavoidable in the proxy pattern.

---

## Summary

**Must Fix (Priority 2):**
- ✅ Fix retry counter race condition
- ✅ Add token refresh rate limiting  
- ✅ Address recursive retry potential

**Optional (Priority 3):**
- Extract HTML templates
- Reduce `any` usage where possible

Once Priority 2 issues are addressed, this PR will be ready to merge. Great work overall! 🎉

StuartFenton:
I don't believe there's a race condition here. Let me explain why:

**The operations are idempotent by design:**

- **Different workspace IDs:** These are independent operations that should run separately - not a race condition, just parallel processing.

- **Same workspace ID:** Multiple concurrent 401 errors all use the same refresh token, receive the same new tokens from Linear, and save them to the same workspace configuration. Running this 5 times simultaneously just writes the same values 5 times - that's idempotent.

The Proxy wrapper ensures that when any call refreshes the token, all subsequent calls benefit from the updated client. Whether one refresh happens or three happen concurrently for the same workspace, the end state is identical.

**Regarding rate limiting:** Rate limiting isn't necessary here - these are independent instances, and unless Linear itself has rate limits we need to respect, the number of simultaneous refresh attempts in practice is minimal. Unless we're running huge numbers of parallel processes, again, rate limits are not necessary.

Connoropolous:
thanks for the reply @StuartFenton 
I'll take a look

StuartFenton:
I dug into the issue and found that Linear's refresh tokens are unique per request. When multiple Linear API calls fail with 401 simultaneously (e.g., token expired), each one independently attempts to refresh the OAuth token. This does indeed create a race condition:

```
Request A gets 401 → refreshes → gets new token T1, refresh token R1
Request B gets 401 → refreshes → gets new token T2, refresh token R2
Request C gets 401 → refreshes → gets new token T3, refresh token R3
```

**The issue:** Linear's refresh tokens are single-use. When Linear issues R2, it invalidates R1. When it issues R3, it invalidates R2.

The result is unpredictable - whichever request saves last "wins", but the intervening refresh tokens have already been invalidated by Linear. This can leave the system in an inconsistent state where the stored refresh token has already been consumed.

## Proposed Solution: Promise Coalescing

Instead of each request performing its own refresh, we coalesce concurrent refresh attempts into a single operation leveraging promises into a queue. This eliminates simultaneous requests and changes them to one after the other.

```typescript
private pendingRefreshes = new Map<string, Promise<{ success: boolean; newToken?: string }>>();

async refreshLinearToken(repositoryId: string): Promise<{ success: boolean; newToken?: string }> {
  const repo = this.repositories.get(repositoryId);
  const workspaceId = repo?.linearWorkspaceId;

  // If refresh already in progress for this workspace, wait for it
  const pending = this.pendingRefreshes.get(workspaceId);
  if (pending) {
    return pending;
  }

  // Start new refresh, store the promise
  const refreshPromise = this.doRefresh(repositoryId);
  this.pendingRefreshes.set(workspaceId, refreshPromise);

  try {
    return await refreshPromise;
  } finally {
    this.pendingRefreshes.delete(workspaceId);
  }
}
```

## How It Works

```
t0:  Request A gets 401
     → calls refreshLinearToken()
     → no pending promise exists
     → creates promise, starts HTTP request to Linear

t1:  Request B gets 401 (A's request still in flight)
     → calls refreshLinearToken()
     → pending promise EXISTS
     → returns same promise, waits for A's result

t2:  Request C gets 401 (A's request still in flight)
     → calls refreshLinearToken()
     → pending promise EXISTS
     → returns same promise, waits for A's result

t3:  A's HTTP request completes with new token
     → A, B, and C all receive the same result
     → all three retry their original API call with the new token
```

**One refresh, three beneficiaries.** The refresh token is used exactly once.

This approach is simple (just a Map of pending promises), requires no external dependencies, and guarantees only one refresh per workspace at a time while allowing concurrent requests to share a single HTTP round-trip.

Connoropolous:
@StuartFenton 
based on your DM, I'm going to wait for your signal having tested this out again over 24-48 hours before we move to merge

StuartFenton:
Implemented the promise coalescing fix discussed above.

**Changes:**
- Removed retry counter and 1-second delay from `wrapLinearClient` (no longer needed)
- Added `pendingRefreshes` Map to coalesce concurrent refresh requests per workspace
- Added `doTokenRefresh` to handle the actual HTTP refresh
- Added tests for concurrent token refresh scenarios

selfhosting/cyrus-agent-auth.sh
Connoropolous, line 1: instead of being in a bash script it'd make more sense for this to be part of the `cli`

Connoropolous (changes requested):
one other thing is that there is too much use of 'any' in the PR for us to merge yet, so that needs to be addressed first.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -5278,4 +5303,259 @@ ${input.userComment}
...
+	 * This is the stable pointer - it never changes. On every call, it looks up
+	 * the current client from rawLinearClients Map. When token refreshes, we just
+	 * update the Map, and all existing references automatically use the new client.
+	 *
+	 * Uses promise coalescing in refreshLinearToken to handle concurrent
+	 * 401 errors - all concurrent callers wait for a single refresh.
+	 */
+	private createLinearClientProxy(repositoryId: string): LinearClient {
+		// Create a proxy that always looks up the current client
+		return new Proxy({} as LinearClient, {
```
Connoropolous: I hadn't seen Proxy before. I'm understanding the use case for it, but I think it probably points at an architecture pattern flaw we may have the fact that this is required? 

Maybe we should be thinking about a SelfRefreshingLinearClient class?

Connoropolous: Thinking more, it seems plausible that it's best to encapsulate it inside of the LinearIssueTracker , since that should be a standalone IIssueTrackerService

StuartFenton:
@Connoropolous Yes, maintainer edits are enabled on the branch - you should be able to push directly. Feel free to resolve conflicts or make changes as needed.

StuartFenton:
@Connoropolous I'll work on reducing the `any` usage in the proxy wrapper code and push an update.

StuartFenton:
## Architectural Refactor: Moving Token Refresh into LinearIssueTrackerService

### Background

In order to encapsulate the token refresh logic inside `LinearIssueTrackerService` rather than using a JavaScript Proxy wrapper in EdgeWorker, we need to address several constraints. This document outlines the proposed approach and its implications.

### The Problem We're Solving

Linear OAuth tokens expire every 24 hours. When a token expires mid-operation, we get a 401 error. The refresh mechanism must handle three critical requirements:

1. **Race condition prevention**: Multiple concurrent API calls may all receive 401s simultaneously. If each independently refreshes the token, Linear's single-use refresh tokens cause chaos - each refresh invalidates the previous one, leaving the system in an inconsistent state.

2. **One refresh updates all**: Multiple repositories can share the same Linear workspace. When we refresh the token for one, all services using that workspace must receive the new token.

3. **Mutable client reference**: The `@linear/sdk` `LinearClient` is immutable - the token is set at construction and cannot be changed. Our current `LinearIssueTrackerService` stores the client as `private readonly linearClient`, so even after a refresh, it would continue using the stale token.

### Proposed Solution

The simplest approach: pass a token provider callback to `LinearIssueTrackerService`. When a 401 occurs, the service requests a fresh token, creates a new client, and retries.

#### Changes to `LinearIssueTrackerService`

1. **Remove `readonly` from `linearClient`** - The `@linear/sdk` `LinearClient` is immutable. There is no `setToken()` or `updateAccessToken()` method - the token is baked into the instance at construction and cannot be changed. This means when we receive a fresh token, we must create an entirely new `LinearClient` instance and replace our reference to it.

   **Potential consequences of client replacement:**
   - Any internal state the `LinearClient` maintains (caching, request queuing, logging configuration, etc.) will be lost
   - Any in-flight requests on the old client may behave unpredictably
   - If the SDK maintains connection pools or other resources, they may not be properly cleaned up

   We should investigate whether `LinearClient` maintains any such state before implementing.

2. **Add optional token provider callback in constructor**:
   ```typescript
   constructor(
     linearClient: LinearClient,
     private refreshToken?: () => Promise<string>
   ) {
     this.linearClient = linearClient;
   }
   ```

3. **Add `withRefresh()` wrapper** - Catches 401 errors, gets fresh token, creates new client, retries:
   ```typescript
   private async withRefresh<T>(fn: () => Promise<T>): Promise<T> {
     try {
       return await fn();
     } catch (error) {
       if (\!this.isTokenExpiredError(error) || \!this.refreshToken) {
         throw error;
       }
       const newToken = await this.refreshToken();
       this.linearClient = new LinearClient({ accessToken: newToken });
       return await fn();
     }
   }
   ```

4. **Wrap all public methods**:
   ```typescript
   async fetchIssue(idOrIdentifier: string): Promise<Issue> {
     return this.withRefresh(() => this.linearClient.issue(idOrIdentifier));
   }
   ```

#### Changes to EdgeWorker

1. **Keep the coalescing logic** - The `pendingRefreshes` Map stays in EdgeWorker. It coordinates across all services sharing a workspace.

2. **Remove Proxy wrapper** - Delete `createLinearClientProxy()` and `rawLinearClients` Map

3. **Pass token provider to each service**:
   ```typescript
   const issueTracker = new LinearIssueTrackerService(
     new LinearClient({ accessToken: repo.linearToken }),
     () => this.refreshTokenForWorkspace(repo.linearWorkspaceId)
   );
   ```

### How Coalescing Works

When multiple services hit 401s simultaneously:

```
Service 1 gets 401 → calls refreshToken() → EdgeWorker creates promise, starts HTTP refresh
Service 2 gets 401 → calls refreshToken() → EdgeWorker returns same pending promise
Service 3 gets 401 → calls refreshToken() → EdgeWorker returns same pending promise

HTTP refresh completes → all three services receive same token string
Each service creates its own new LinearClient with that token
Each service retries its original operation
```

All three services share the same promise. One HTTP request, one token, no race condition.

### Why Return Token Instead of Client

- **Simpler contract** - The callback's job is just "provide a valid token"
- **Service owns its client lifecycle** - Creating `LinearClient` is the service's responsibility
- **EdgeWorker stays decoupled** - It returns data (a token string), not Linear-specific objects
- **Clean separation** - The shared promise contains just the token; each service decides what to do with it

### What This Removes

- `createLinearClientProxy()` method
- `rawLinearClients` Map
- All Proxy-related `any` types

### Migration Path

1. Investigate `LinearClient` internals to understand impact of instance replacement
2. Modify `LinearIssueTrackerService` (add callback, `withRefresh`, wrap methods)
3. Update EdgeWorker to pass token provider callback and remove Proxy code
4. Update tests

StuartFenton:
**Addendum to the above:**

Regarding the "potential consequences of client replacement" - the current Proxy approach also creates a new `LinearClient` instance on token refresh:

```typescript
// In doTokenRefresh() - lines 5536-5539
const newClient = new LinearClient({
    accessToken: data.access_token,
});
this.rawLinearClients.set(repoId, newClient);
```

So we're already doing client replacement - we're just changing where it happens (from EdgeWorker's Map to LinearIssueTrackerService's internal reference). The potential consequences apply equally to both approaches, but worth noting that this isn't a new risk we'd be introducing.

Connoropolous:
right... 

"Potential consequences of client replacement:

Any internal state the LinearClient maintains (caching, request queuing, logging configuration, etc.) will be lost
Any in-flight requests on the old client may behave unpredictably
If the SDK maintains connection pools or other resources, they may not be properly cleaned up
We should investigate whether LinearClient maintains any such state before implementing."

I am thinking there is likely little value add in the persistent LinearClient. 

if we can quickly confirm this then we could proceed with an approach where the LinearIssueTrackerService, whenever it needs a LinearClient, constructs it "on the spot" having ensured token validity and refreshed beforehand.

StuartFenton:
**Follow-up: Simpler Solution Found**

It didn't make sense to me that replacing the entire client was the only way to handle this. So I spent more time and dug through the source code of Linear's SDK, tracing through how requests are made, and found there's actually a `setHeader(key, value)` method on the underlying `LinearGraphQLClient`.

The `LinearClient` exposes its internal GraphQL client via `linearClient.client`, and each request reads from `this.options.headers` fresh - the token isn't cached anywhere else.

So instead of replacing the entire client:

```typescript
// What I originally proposed - creates new client, loses state
const newToken = await this.refreshToken();
this.linearClient = new LinearClient({ accessToken: newToken });
```

We can just update the Authorization header:

```typescript
// Much simpler - same client, just update the header
const newToken = await this.refreshToken();
this.linearClient.client.setHeader('Authorization', `Bearer ${newToken}`);
```

**This means:**
- No need to replace the client instance
- `readonly` can stay on `linearClient` 
- No lost internal state (connections, caching, etc.)
- One line to update the token

The `withRefresh()` implementation becomes:

```typescript
private async withRefresh<T>(fn: () => Promise<T>): Promise<T> {
  try {
    return await fn();
  } catch (error) {
    if (\!this.isTokenExpiredError(error) || \!this.refreshToken) {
      throw error;
    }
    const newToken = await this.refreshToken();
    this.linearClient.client.setHeader('Authorization', `Bearer ${newToken}`);
    return await fn();
  }
}
```

I haven't tested this yet, but in theory this is maybe what they intended for this use case - the [Advanced usage documentation](https://linear.app/developers/advanced-usage) mentions `setHeader` for custom headers, but doesn't explicitly cover using it for token refresh. The OAuth documentation explains how to get new tokens but not how to apply them to an existing client.

StuartFenton:
**Update: Tested and confirmed working ✅**

```
=== Testing LinearClient setHeader ===

1. Initial state:
   client.client exists: true
   client.client.setHeader exists: function
   Initial Authorization header: Bearer initial_token_123

2. Calling setHeader('Authorization', 'Bearer new_token_456')...
   Updated Authorization header: Bearer new_token_456

3. Result: ✅ SUCCESS - Header was updated!
```

The `setHeader` approach works. We can update the Authorization header in place without replacing the client.

Connoropolous:
This is good progress, simpler!

From an abstractions and code organization POV I am still keen to extract token refresh from EdgeWorker and have it encapsulated within the LinearIssueTrackerService since its an implementation detail of that service.

StuartFenton:
## Token Refresh Refactor

Moved token refresh logic from EdgeWorker to LinearIssueTrackerService to properly encapsulate it as an implementation detail of the Linear service.

**What's now in LinearIssueTrackerService (packages/linear-event-transport):**
- Token refresh HTTP calls to Linear OAuth endpoint
- Workspace-level coalescing via static `pendingRefreshes` map (prevents race conditions)
- Shared refresh token state via static `workspaceRefreshTokens` map (all instances share the same token)
- Per-instance LinearClient header updates

**What remains in EdgeWorker (minimal, cannot be fully removed):**
- Passes initial credentials via `LinearOAuthConfig`
- `onTokenRefresh` callback for persistence only

**Why EdgeWorker still has code:** EdgeWorker owns `config.json`, so it needs to persist refreshed tokens to disk. The `onTokenRefresh` callback is the minimum interface needed - LinearIssueTrackerService handles all refresh logic and just notifies EdgeWorker when tokens change so they can be saved.

---

I'm going to run this for a few days to make sure there are no issues with the token refresh flow before merging.

TyceHerrman:
@StuartFenton have you been able to confirm no issues with token refresh? Looking forward to unblocked self-hosting!

Connoropolous:
@TyceHerrman 
Stuart opened up a new version of this PR, here: https://github.com/ceedaragents/cyrus/pull/656#pullrequestreview-3606504164

It's looking good to me, Stuart said he's been running for days, and it also has the `self-add-repo` command you'd need for adding repository configs!

In case you haven't yet, join the Discord btw!
https://discord.gg/zGvqJMuG

PR #807, opened 2026-01-26
feat: add routing labels when adding repos

StuartFenton:
## Problem

When adding a repository with `cyrus self-add-repo`, the command doesn't set `routingLabels`. This means when you create a Linear issue with a label matching your repo name, the RepositoryRouter can't match it via label-based routing (Priority 1).

Instead, it falls through to catch-all routing (Priority 4), which selects the first repository with no routing configuration. This often picks the wrong repository.

## Solution

- Always add `routingLabels` when creating a repository config
- Default to the repository name (e.g., `routingLabels: ["myrepo"]`)
- Allow custom labels via `-l` flag for cases where the Linear label differs from the repo name

## Usage

```bash
# Uses repo name as routing label (default)
cyrus self-add-repo https://github.com/org/myrepo
# → routingLabels: ["myrepo"]

# Custom routing labels
cyrus self-add-repo https://github.com/org/myrepo -l backend,api
# → routingLabels: ["backend", "api"]
```

## Test plan

- [x] All existing tests pass (56 tests)
- [x] Added test for custom `-l` flag
- [x] Typecheck passes

### DIFF
--- a/apps/cli/src/app.ts
+++ b/apps/cli/src/app.ts
@@ -108,9 +108,9 @@ program
 
 // Self-add-repo command - Clone and add repository
 program
-	.command("self-add-repo [url] [workspace]")
+	.command("self-add-repo <url> [workspace]")
 	.description(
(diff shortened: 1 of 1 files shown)


fuxialexander:
Very useful!!! Hopefully it can be integrated soon!

PR #868, opened 2026-02-17
feat: Slack sessions use empty working directories, not git worktrees (CYPACK-815)

cyrusagent:
## Summary

Decouples Slack agent sessions from repository configuration so they run in transient empty directories rather than git worktrees. Subsequent `@mentions` in the same Slack thread now share the same session context.

Resolves [CYPACK-815](https://linear.app/ceedar/issue/CYPACK-815).

## Changes

- **Empty working directories for Slack**: `createWorkspace` creates plain directories under `~/.cyrus/slack-workspaces/` keyed by thread (`channel:thread_ts`), instead of creating git worktrees off `origin/main`
- **No repository dependency**: Slack webhook handler no longer picks the first configured repository. A dedicated `AgentSessionManager` with `NoopActivitySink` manages Slack sessions independently
- **Thread-based session sharing**: A `threadSessions` map tracks `thread_key → session_id`. Follow-up `@mentions` in the same thread either inject via `addStreamMessage` (if session is running) or resume via `resumeSessionId` (if session completed)
- **Issue-free sessions**: Added `AgentSessionManager.createChatSession()` for chat platforms that don't have issue context, replacing the use of `createLinearAgentSession` with synthetic `IssueMinimal` objects
- **`ChatSessionHandler` + `ChatPlatformAdapter`**: Generic session lifecycle engine extracted from EdgeWorker, with `SlackChatAdapter` as the first platform implementation
- **New `NoopActivitySink`**: A no-op implementation of `IActivitySink` for platforms that don't post activities to an external tracker
- **New `SlackReactionService`**: Service for adding emoji reactions to Slack messages (was missing from the `cyrus-slack-event-transport` package)

## Test plan

- [x] All 508 edge-worker tests pass
- [x] All package tests pass (core, claude-runner, slack-event-transport, github-event-transport, etc.)
- [x] TypeScript typecheck clean across all 12 packages
- [x] Build succeeds
- [x] Linting clean (0 errors, 17 pre-existing warnings in unrelated files)

### DIFF
--- a/packages/edge-worker/src/ChatSessionHandler.ts
+++ b/packages/edge-worker/src/ChatSessionHandler.ts
@@ -124,13 +124,16 @@ export class ChatSessionHandler<TEvent> {
 
 				if (existingSession && existingRunner?.isRunning()) {
 					// Session is actively running — inject the follow-up via streaming input
-					if (existingRunner.addStreamMessage) {
+					if (
+						existingRunner.addStreamMessage &&
+						existingRunner.isStreaming?.()
+					) {
 						this.logger.info(
 							`Injecting follow-up prompt into running session ${existingSessionId} (thread ${threadKey})`,
 						);
 						existingRunner.addStreamMessage(taskInstructions);
 					} else {
-						// Runner doesn't support streaming input — notify user
+						// Runner doesn't support streaming input or isn't in streaming mode — notify user
 						this.logger.info(
 							`Session ${existingSessionId} is still running, notifying user (thread ${threadKey})`,
 						);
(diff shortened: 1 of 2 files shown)


packages/edge-worker/src/ChatSessionHandler.ts
```diff
@@ -0,0 +1,427 @@
...
+
+			// Create an internal agent session (not tied to any repository)
+			const sessionId = `${this.adapter.platformName}-${eventId}`;
+			this.sessionManager.createLinearAgentSession(
+				sessionId,
+				sessionKey,
+				issueMinimal,
+				workspace,
+				this.adapter.platformName,
+			);
```
PaytonWebber: Why is this still called `createLinearAgentSession`? These names should be abstracted, since agent sessions are no longer always linked to Linear.

packages/edge-worker/src/ChatSessionHandler.ts
```diff
@@ -0,0 +1,427 @@
...
+			const eventId = this.adapter.getEventId(event);
+			const sessionKey = `${this.adapter.platformName}:${threadKey}`;
+			const issueMinimal: IssueMinimal = {
+				id: sessionKey,
+				identifier: `${this.adapter.platformName}-${eventId}`,
+				title:
+					taskInstructions.slice(0, 100) +
+					(taskInstructions.length > 100 ? "..." : ""),
+				branchName: "",
+			};
```
PaytonWebber: I don't like that this is trying to match the format of a Linear issue. Sessions should not depend on this issue context.

PR #902, opened 2026-02-25
feat: reduce Linear API noise, auto-retry on billing limit

jensenbox:
## Problem

Two issues compound the rate-limit problem addressed in the earlier PRs in this stack:

**1. Read-only tool noise**  
Every `Read`, `Glob`, `Grep`, `WebFetch`, and `WebSearch` call posts a Linear activity entry. A typical coding session makes dozens of these per task. At scale this generates hundreds of unnecessary Linear API calls per session and clutters the issue activity feed with implementation noise that reviewers don't care about.

**2. Billing limit hard-stops**  
When Anthropic's daily billing limit is hit (distinct from the per-minute rate limit), the session ends with a `billing_error` SDK message. Previously the session was marked as permanently failed even though billing limits reset on a predictable UTC schedule. Users had to manually re-trigger their tickets the next day.

**3. Misleading error badges**  
Transient events (rate limits, billing limits) were posted to Linear as `error` type, setting a red error badge on the issue. These are normal operational events, not bugs.

### Activity sink filtering

Skip Linear activity posting for read-only tool results when they succeed. Tools filtered: `Read`, `Glob`, `Grep`, `WebFetch`, `WebSearch`, `LS`. Errors from these tools are still surfaced. The corresponding tool-call-setup entry is also suppressed to avoid orphaned "called tool X" entries.

**Before:** ~200 Linear API calls/session  
**After:** ~60–80 Linear API calls/session (reads account for ~65% of tool volume in typical tasks)

### Auto-retry after billing limit reset

When a `billing_error` SDK message is received:
- Parse the reset time from the error message (e.g. `"resets 6am (UTC)"`)
- Schedule a `setTimeout` to automatically resume the session after reset + 2-minute buffer
- Post an informational `thought` to the Linear issue so users know what's happening

```
INFO Billing limit hit — will auto-retry at 06:02 UTC
```

### Transient errors as thoughts

`billing_error` and `rate_limit` SDK errors now post to Linear as `thought` type rather than `error` type, avoiding the red error badge for events that resolve automatically.

## Changes

- `packages/core/src/CyrusAgentSession.ts` — `billingError` metadata field
- `packages/edge-worker/src/AgentSessionManager.ts`
  - `postThoughtToLinear()` helper
  - `billingLimitHit` event + `billing_error` handling
  - Read-only tool filtering in activity sink
  - Transient SDK errors posted as thoughts
- `packages/edge-worker/src/EdgeWorker.ts`
  - `billingRetryTimeouts` map + `parseBillingResetDelayMs()`
  - Subscribe to `billingLimitHit`, schedule retry, clean up on `stop()`
  - Wire `maxConcurrentSessions` through config hot-reload path
- `packages/edge-worker/src/ConfigManager.ts` — `maxConcurrentSessions` passthrough
- `apps/cli/src/services/WorkerService.ts` — pass `maxConcurrentSessions` to `EdgeWorker`

## Stack

Builds on #865, #900, and #901. Completes the rate-limit resilience series:

| # | PR | Description |
|---|-----|-------------|
| 1 | #865 | Linear API 429 retry-after handling |
| 2 | #900 | Claude SDK `rate_limit_event` handler |
| 3 | #901 | `maxConcurrentSessions` FIFO queue |
| **4** | **This PR** | **Activity sink filtering + billing auto-retry** |

Full stack with incremental diffs: [closient/cyrus](https://github.com/closient/cyrus/pulls)

### DIFF
--- a/packages/core/src/CyrusAgentSession.ts
+++ b/packages/core/src/CyrusAgentSession.ts
@@ -73,6 +73,16 @@ export interface CyrusAgentSession {
 		totalCostUsd?: number;
 		usage?: any;
 		commentId?: string;
+		/**
+		 * Set when an Anthropic billing/usage-limit error is detected mid-session.
+		 * EdgeWorker uses this to schedule an automatic retry after the API resets
+		 * rather than marking the session as permanently failed.
+		 */
+		billingError?: {
+			/** Raw error message from the SDK, e.g. "You've hit your limit · resets 6am (UTC)" */
+			message: string;
+			detectedAt: number;
+		};
 		procedure?: {
 			procedureName: string;
 			currentSubroutineIndex: number;
--- a/packages/edge-worker/src/AgentSessionManager.ts
+++ b/packages/edge-worker/src/AgentSessionManager.ts
@@ -67,6 +67,16 @@ export interface AgentSessionManagerEvents {
 		/** Current iteration (1-based) */
 		iteration: number;
 	}) => void;
+	/**
+	 * Emitted when an Anthropic billing/usage-limit error ends a session.
+	 * EdgeWorker responds by scheduling an automatic retry after the reset time.
+	 */
+	billingLimitHit: (data: {
+		sessionId: string;
+		session: CyrusAgentSession;
+		/** Raw reset message from the SDK, e.g. "You've hit your limit · resets 6am (UTC)" */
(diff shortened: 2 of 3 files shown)


Connoropolous:
thanks for the PRs @jensenbox ! 
We will review them ourselves, asap!

PR #905, opened 2026-02-26
fix: recover from missing session/repository mapping after restart (CYPACK-852)

cyrusagent:
Assignee: [payton](https://linear.app/ceedar/profiles/payton)

## Summary

Fixes a bug where Linear sessions appeared stuck/hung when prompted, stop, unassignment, or issue update webhooks arrived for sessions whose in-memory cache mappings were lost after a restart or migration. Previously, 4 EdgeWorker webhook handlers returned silently on cache miss, providing no user feedback.

**Issue:** [CYPACK-852](https://linear.app/ceedar/issue/CYPACK-852)

### Prompted webhook (Branch 3) — `handleUserPromptedAgentActivity`
Added 3-tier fallback when `issueRepositoryCache` misses:
1. Search all `agentSessionManagers` for the session by ID
2. Re-route via `RepositoryRouter.determineRepositoryForWebhook()`
3. Post a visible error activity if all recovery fails

On successful recovery, the cache mapping is re-established for future webhooks.

### Stop signal — `handleStopSignal`
When no session is found across any manager, now posts an acknowledgment response activity instead of returning silently, so users don't see a hanging state.

### Unassignment — `handleIssueUnassignedWebhook`
When repository cache misses, searches all `agentSessionManagers` for sessions matching the issue ID and proceeds with normal unassignment flow.

### Issue update — `handleIssueContentUpdate`
Same fallback pattern as unassignment — searches all managers for sessions matching the issue before giving up.

### Log level adjustments
- Recovery attempts logged at `info` level (expected legacy cases)
- Only true failures (all recovery exhausted) use `warn`
- `debug` for cases where no sessions exist at all

## Testing

- 8 new tests in `EdgeWorker.missing-session-recovery.test.ts` covering all 4 handlers
- Full test suite passes: 537 tests across 43 files
- Linting clean, TypeScript types valid across all 15 packages

### DIFF
--- a/packages/edge-worker/src/EdgeWorker.ts
+++ b/packages/edge-worker/src/EdgeWorker.ts
@@ -3477,8 +3477,7 @@ ${taskInstructions}
 					const repos = Array.from(this.repositories.values());
 					const routingResult =
 						await this.repositoryRouter.determineRepositoryForWebhook(
-							// Both webhook types share the same underlying Linear SDK type
-							webhook as unknown as AgentSessionCreatedWebhook,
+							webhook,
 							repos,
 						);
 
@@ -3507,7 +3506,7 @@ ${taskInstructions}
 				if (firstManager) {
 					await firstManager.createResponseActivity(
 						agentSessionId,
-						"I couldn't process your message because the session configuration was lost after a restart. Please create a new session by re-assigning the issue.",
+						"I couldn't process your message because the session configuration was lost. Please create a new session by mentioning me (@cyrus) in a new comment with your prompt.",
 					);
 				}
 				this.logger.warn(
--- a/packages/edge-worker/src/RepositoryRouter.ts
+++ b/packages/edge-worker/src/RepositoryRouter.ts
@@ -128,7 +128,7 @@ export class RepositoryRouter {
 	 * Priority 5: Catch-all repositories
 	 */
 	async determineRepositoryForWebhook(
-		webhook: AgentSessionCreatedWebhook,
+		webhook: AgentSessionCreatedWebhook | AgentSessionPromptedWebhook,
 		repos: RepositoryConfig[],
 	): Promise<RepositoryRoutingResult> {
 		const workspaceId = webhook.organizationId;


packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -3398,12 +3447,74 @@ ${taskInstructions}
...
+
+			if (!repository) {
+				// All recovery attempts failed - post visible feedback
+				const firstManager = this.agentSessionManagers.values().next().value as
+					| AgentSessionManager
+					| undefined;
+				if (firstManager) {
+					await firstManager.createResponseActivity(
+						agentSessionId,
+						"I couldn't process your message because the session configuration was lost after a restart. Please create a new session by re-assigning the issue.",
```
PaytonWebber: Change this message to: "I couldn't process your message because the session configuration was lost. Please create a new session by mentioning me (@cyrus) in a new comment with your prompt."

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -3398,12 +3447,74 @@ ${taskInstructions}
...
+			}
+
+			// Second fallback: re-route via repository router
+			if (!repository) {
+				try {
+					const repos = Array.from(this.repositories.values());
+					const routingResult =
+						await this.repositoryRouter.determineRepositoryForWebhook(
+							// Both webhook types share the same underlying Linear SDK type
+							webhook as unknown as AgentSessionCreatedWebhook,
```
PaytonWebber: This looks risky. You are not supposed to use `as unknown as Type` patterns. The better approach would be to update the `determineRepositoryForWebhook` method to use a discriminated union, that accepts `AgentSessionPromptedWebhook` or `AgentSessionCreatedWebhook`

PaytonWebber:
@cyrusagent address the change requests I left on the PR

ghost:
This repository is not configured with Cyrus. Please add it at https://app.atcyrus.com/repos to start using Cyrus here.

PR #940, opened 2026-03-05
refactor: replace hardcoded runner instantiation with factory pattern

PaytonWebber:
## Summary

- **Decouple `ChatSessionHandler` from `ClaudeRunner`** by introducing a `createRunner` factory in `ChatSessionHandlerDeps`, enabling any runner type (Claude, Gemini, Codex, Cursor) for chat sessions
- **Consolidate runner instantiation** into a single `createRunnerForType()` method in `EdgeWorker` with an exhaustive `switch` (compile-time safety via `satisfies never`)
- **Fix model/runner-type mismatch on config hot-reload** — model resolution now happens inside the factory closure alongside runner type resolution, so the two always stay in sync
- **Fix unhandled promise rejection** in `onMessage` callback — now returns the promise instead of fire-and-forget
- **Add `logger` field to `AgentRunnerConfig`** to formalize existing usage across runners
- **Include agent's own messages in Slack thread context** — previously the bot's replies were filtered out, so follow-up sessions (especially after a runner type change) lost conversation history. Now all messages are included with the agent's own labeled as `assistant (you)`

## Test plan

- [x] `pnpm typecheck` passes across all packages
- [x] `pnpm test:packages:run` — all 537 tests pass
- [x] Pre-commit hooks (biome lint + typecheck) pass

### DIFF
--- a/packages/edge-worker/src/EdgeWorker.ts
+++ b/packages/edge-worker/src/EdgeWorker.ts
@@ -32,6 +32,7 @@ import type {
 	IssueUnassignedWebhook,
 	IssueUpdateWebhook,
 	RepositoryConfig,
+	RunnerType,
 	SerializableEdgeWorkerState,
 	SerializedCyrusAgentSession,
 	SerializedCyrusAgentSessionEntry,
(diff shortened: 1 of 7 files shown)


packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -3788,6 +3787,27 @@ ${taskSection}`;
 		);
 	}
 
+	/**
+	 * Instantiate the appropriate runner for the given type.
+	 */
+	private createRunnerForType(
+		runnerType: "claude" | "gemini" | "codex" | "cursor",
```
Connoropolous, line 3791: @cyrusagent isn't there an enum or discrimated union somehwere that we can refer to for these magic strings? can you replace, if so.

PR #952, opened 2026-03-09
Start cloudflare tunnel when running self-auth

bertbijnens:
Start cloudflare tunnel when running self-auth.

This way Cyrus can run completely isolated inside docker without having to change the CYRUS_BASE_URL config for running self-auth later on.

### DIFF
--- a/apps/cli/src/commands/SelfAuthCommand.ts
+++ b/apps/cli/src/commands/SelfAuthCommand.ts
@@ -54,6 +54,20 @@ export class SelfAuthCommand extends BaseCommand {
 		console.log();
 
 		try {
+			if (process.env.CLOUDFLARE_TOKEN) {
+				this.logger.info("Starting cloudflare tunnel...");
+
+				const { SharedApplicationServer } = await import("cyrus-edge-worker");
+				const sharedApplicationServer = new SharedApplicationServer(
+					this.callbackPort,
+					baseUrl,
+					false,
+				);
+				await sharedApplicationServer.startCloudflareTunnel(
+					process.env.CLOUDFLARE_TOKEN,
+				);
+			}
+
 			// Start temporary server to receive OAuth callback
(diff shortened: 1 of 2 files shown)


Connoropolous:
Thanks @bertbijnens ! 
Can you clarify the before/after for this PR? I know you kind of did, but putting a finer point on it?

bertbijnens:
@Connoropolous Yes of course!

Before:
When you run `clyne self-auth` it starts a Fastify webserver with one route `/callback `.
This doesn't use the same webserver as `clyne` runs on and will not start the Cloudflare tunnel.
In case the user doesn't use Cloudflare there is no problem. The webserver runs on the same host as the user specified in `CYRUS_BASE_URL`. The user can successfully authenticate with linear and the callback is reachable.

However, when the user uses a Cloudflare tunnel, the `CYRUS_BASE_URL` will point to that Cloudflare domain and without the tunnel being active, the `clyne self-auth Fastify webserver` is not accessible. 
A workaround solution would be to temporarily set the `CYRUS_BASE_URL` to localhost, run self auth and then revert it back to the Cloudflare tunnel domain.


Solution:
Start the Cloudflare tunnel while running `clyne self-auth`. This way the env config doesn't need to change, and `/callback` is reachable through Cloudflare.

After:
Inside the `clyne self-auth` command we verify if the user has a `CLOUDFLARE_TOKEN` token configured.
In case this is set we start the Cloudflare tunnel and wait for the connection to succeed.
We then continue to start the Fastify webserver as before.

Inside `packages/edge-worker/src/SharedApplicationServer.ts` I changed the private `startCloudflareTunnel` method to public.

Result:
Users with our without Cloudflare tunnels, can use the self-auth command to reconnect or connect their Linear. Without having to change the config.

Hope this clarifies it!

Connoropolous:
that makes sense. Thanks for contributing!
Merged.

PR #993, opened 2026-03-16
fix: prevent pipeline hang when subroutine completes with empty result

acarr:
## Problem

Cyrus intermittently appears "hung" in Linear — the timer keeps running, the transcript shows work is done, but the ticket is never updated with a commit, PR, or summary. The only recourse is to manually stop the session.

## Root Cause

In `AgentSessionManager.completeSession()`, the condition that decides whether to advance the subroutine pipeline is:

```js
if ("result" in resultMessage && resultMessage.result) {
```

Some subroutines (observed on `changelog-update`, but could affect any) complete with `subtype: "success"` but **no result text** — `result` is `undefined` or empty string. When this happens:

1. The `if` above evaluates to `false` (result is falsy)
2. All `else if` branches check `subtype !== "success"`, which is also `false`
3. The function **silently returns** — no pipeline advancement, no error, no log
4. The Linear ticket shows "still working" indefinitely

## Fix

Change the primary check from requiring result text to checking the success status:

```js
if (resultMessage.subtype === "success") {
```

When result text is empty, a placeholder `"(completed successfully)"` is synthesized so downstream code that expects a string still works.

Also fixes a secondary issue: when a user clicks "stop" in Linear but the session had already completed successfully, the pipeline was being abandoned. Now it checks whether the stop was actually effective (non-success result) vs. arrived too late (success result).

### Reproduced
- **ABC-215, ABC-216**: Both hung at `changelog-update` with diagnostic logs showing `subtype=success hasResult=false` followed by silence
- **ABC-227, ABC-231**: Hit the same `hasResult=false` condition, but with the fix active — logs show `success with empty result — synthesizing result text` and the pipeline continued to completion

### Diagnostic log showing the fix in action (WAZ-227)
```
completeSession ENTERED ... subtype=success hasResult=true    ← normal subroutine
completeSession ENTERED ... subtype=success hasResult=true    ← normal subroutine
completeSession ENTERED ... subtype=success hasResult=false   ← BUG TRIGGERED
  success with empty result — synthesizing result text        ← FIX CAUGHT IT
  calling handleProcedureCompletion                           ← pipeline continued
completeSession ENTERED ... subtype=success hasResult=true    ← rest completed normally
```

### Historical analysis
Analyzed 28 previous Cyrus tickets. 6 hung with the same pattern — `Session completed with N messages` logged by ClaudeRunner but no `Subroutine completed, advancing to next` from AgentSessionManager. Task type and complexity showed no correlation — both simple (Settings screen) and complex (4-package polling system) tickets hung.

## Test plan

- [x] New test: `advances pipeline when success result has empty result text` — the core regression test
- [x] New test: `continues pipeline when stop arrives too late` — stop + success doesn't kill pipeline  
- [x] Updated test: `skips pipeline when stop is effective` — uses non-success subtype
- [x] All 630 existing tests pass
- [x] TypeScript compiles cleanly (typecheck passes in pre-commit hook)
- [x] Verified with 16+ live Cyrus tickets across multiple batches

### DIFF
--- a/packages/edge-worker/src/AgentSessionManager.ts
+++ b/packages/edge-worker/src/AgentSessionManager.ts
@@ -387,7 +387,14 @@ export class AgentSessionManager extends EventEmitter {
 		// to prevent memory leaks
 
 		const wasStopRequested = this.consumeStopRequest(sessionId);
-		const status = wasStopRequested
+
+		// If stop was requested but the session completed successfully, the stop
+		// arrived too late — the runner had already finished its work. In this case,
+		// treat it as a normal completion so the pipeline can continue.
+		const stopWasEffective =
+			wasStopRequested && resultMessage.subtype !== "success";
+
+		const status = stopWasEffective
 			? AgentSessionStatus.Error
 			: resultMessage.subtype === "success"
 				? AgentSessionStatus.Complete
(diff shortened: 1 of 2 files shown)


Connoropolous:
thanks @acarr we are queueing this up for review

PR #1023, opened 2026-03-20
switch cursor-agent cli execution to cursor acp

cyrusagent:
## Summary
- switch Cursor runner execution from the legacy printed CLI stream to ACP JSON-RPC
- add Cursor streaming follow-up handling and ACP startup coverage
- fix F1 CLI-mode regressions needed for label-based Cursor validation and document the passing drive

## Testing
- pnpm --filter cyrus-cursor-runner test:run
- pnpm --filter cyrus-codex-runner test:run
- pnpm --filter cyrus-edge-worker test:run -- EdgeWorker.runner-selection.test.ts
- live F1 drive documented in apps/f1/test-drives/2026-03-19-cursor-acp-label-validation.md

### DIFF
--- a/CHANGELOG.internal.md
+++ b/CHANGELOG.internal.md
@@ -4,6 +4,9 @@ This changelog documents internal development changes, refactors, tooling update
 
 ## [Unreleased]
 
+### Changed
+- Switched `CursorRunner` from the legacy printed CLI stream to ACP JSON-RPC. Added ACP prompt/session handling, ACP tool/event mapping, Cursor streaming follow-up prompt support, ACP startup coverage, and a live F1 validation report. Also fixed F1 CLI-mode setup gaps uncovered during the drive by seeding `linearWorkspaces` in `apps/f1/server.ts` and making `McpConfigService` skip `cyrus-tools` injection when the issue tracker has no Linear client. ([CYPACK-999](https://linear.app/ceedar/issue/CYPACK-999/switch-cursor-agent-cli-execution-to-cursor-acp), [#1023](https://github.com/ceedaragents/cyrus/pull/1023))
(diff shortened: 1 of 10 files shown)


JiwaniZakir (commented):
The `AcpPendingRequest` interface in `CursorRunner.ts` types `resolve` as `(value: unknown) => void`, which means every call site that resolves a pending request will need an unsafe cast or runtime check to use the result — the `AcpInitializeResult` type defined right above it goes unused for that purpose. Narrowing `resolve` to `(value: AcpInitializeResult) => void` (or a discriminated union if other request types are pending) would push the type safety to the boundary where it belongs.

The `linearWorkspaces` entry added to `apps/f1/server.ts` hardcodes `linearToken: "cli-test-token"` directly in the source file. Even though this is a test harness, committing a token-shaped string sets a pattern that's easy to replicate carelessly; an empty string or an env-var fallback (`process.env.LINEAR_TOKEN ?? ""`) would be safer and clearer about intent.

The `AcpInitializeResult.protocolVersion` field is typed as `number`, but ACP/MCP protocol versions are conventionally date strings (e.g. `"2024-11-05"`). If Cursor's ACP ever returns a string there, the field will silently deserialize as `undefined` under strict parsing, which could cause the capability negotiation path to fall back unexpectedly.

cyrusagent:
Superseded by #1169 (CYPACK-1149): switched directly to `@cursor/sdk` instead of going through ACP. The new approach uses .cursor/hooks.json for permission enforcement (validated as the actual gate in headless mode, unlike .cursor/cli.json) and passes mcpServers inline to Agent.create(), eliminating the agent mcp list/enable preflight.

PR #1059, opened 2026-04-01
fix: prevent config watcher feedback loop during OAuth token refresh

QiuYi111:
## Summary

Fixes the OAuth token refresh feedback loop and concurrent token exhaustion that forces users to repeatedly re-authenticate.

Three changes across two files:

### 1. Update in-memory config before writing to disk (`EdgeWorker.saveOAuthTokens`)
When the file watcher fires after `config.json` is updated, it compares in-memory vs on-disk state. By updating in-memory first, no diff is detected, preventing the feedback loop.

### 2. Cooldown on static refresh coalescing (`LinearIssueTrackerService.doTokenRefresh`)
Keep the resolved `pendingRefreshes` promise for 5 seconds instead of clearing immediately in `finally`. Late-arriving 401s coalesce onto the resolved promise instead of triggering redundant HTTP refresh calls that fail because the OAuth refresh token is single-use.

### 3. Clear stale instance promise on retry failure (`LinearIssueTrackerService`)
If the retry with a refreshed token also returns 401, clear the stale `refreshPromise` so the next request can trigger a fresh refresh. Without this, the instance-level coalesce holds a stale promise forever and the token can never be refreshed again after natural expiration (~1 hour).

## Problem

When OAuth tokens are refreshed, `saveOAuthTokens` writes to `config.json` → chokidar detects the change → `handleConfigChange` fires → `detectGlobalConfigChanges` sees `linearWorkspaces` changed → `updateLinearWorkspaceTokens` calls `setAccessToken()` which clears instance-level coalescing → concurrent 401s trigger new refreshes → refresh token already consumed → 400 → token exhaustion → user must re-auth.

## Test Plan

- [x] All 642 existing tests pass (edge-worker)
- [x] `pnpm biome check` passes
- [x] Full monorepo `typecheck` passes
- [x] Tested locally: OAuth token refresh no longer triggers config watcher loop
- [x] Tested locally: ran for hours without needing to re-authenticate (previously required re-auth every ~1 hour)

### DIFF
--- a/packages/edge-worker/src/EdgeWorker.ts
+++ b/packages/edge-worker/src/EdgeWorker.ts
@@ -6625,6 +6625,18 @@ ${input.userComment}
 		}
 
 		try {
+			// Update in-memory config FIRST, then write to disk.
+			// When the file watcher fires, it will see no diff between
+			// in-memory and on-disk config, preventing a feedback loop.
+			const wsConfig = this.config.linearWorkspaces?.[tokens.linearWorkspaceId];
+			if (wsConfig) {
+				wsConfig.linearToken = tokens.linearToken;
+				if (tokens.linearRefreshToken) {
+					wsConfig.linearRefreshToken = tokens.linearRefreshToken;
+				}
+			}
+			this.configManager.setConfig(this.config);
+
 			const configContent = await readFile(this.configPath, "utf-8");
 			const config = JSON.parse(configContent);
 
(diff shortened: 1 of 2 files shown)


Connoropolous:
instead of that, we propose to first change the in memory, then to update the file.

the lack of differences between the in memory and the file will mean no further restarts or cyclical change loop

Connoropolous:
can you give this a try and let us know if that holds up for you?

QiuYi111:
Great suggestion! Updated the approach: we now update in-memory config first via `configManager.setConfig()`, then write to disk. When the file watcher fires, it sees no diff between in-memory and on-disk state, so no feedback loop — without needing to pause/resume the watcher.

This is cleaner and simpler. All 642 tests pass.

PR #1077, opened 2026-04-07
fix: capture OpenCode LLM responses from freeze events (non-streaming models)

frozenbob0306:
Fixes OpenCode runner empty response bug by capturing text from freeze events. See commit message for details.


frozenbob0306:
Sorry.

PR #1112, opened 2026-04-16
fix: update Linear workspace tokens before adding new repositories

cyrusagent:
Assignee: @PaytonWebber ([payton](https://linear.app/ceedar/profiles/payton))

## Summary

Fixes a race condition in the EdgeWorker's `configChanged` handler where new repositories were initialized before Linear workspace tokens were registered. During cloud runtime provisioning, when a new workspace and its first repository arrive in the same config change, the repository initialization fails with "No Linear workspace config found" because `updateLinearWorkspaceTokens()` ran after `addNewRepositories()`.

**Change:** Reorder the handler to call `updateLinearWorkspaceTokens()` before `addNewRepositories()`, ensuring tokens are always available when repositories are initialized.

**Companion PR:** ceedaragents/cyrus-hosted#680

## Test plan

- [x] All 568 edge-worker tests pass
- [x] Full monorepo build and typecheck pass
- [ ] Deploy and verify cloud runtime provisioning succeeds without "No Linear workspace config" errors

[CYPACK-1089](https://linear.app/ceedar/issue/CYPACK-1089)

---
> **Tip:** I will respond to comments that @ mention @cyrusagent on this PR. You can also submit a review with all your feedback at once, and I will automatically wake up to address each comment.

### DIFF
--- a/packages/edge-worker/src/EdgeWorker.ts
+++ b/packages/edge-worker/src/EdgeWorker.ts
@@ -2551,26 +2551,9 @@ ${taskSection}`;
 				// Add to internal map
 				this.repositories.set(repo.id, resolvedRepo);
 
-				// Create issue tracker for this workspace if not already present
-				if (!this.issueTrackers.has(requireLinearWorkspaceId(repo))) {
-					const linearToken = this.getLinearTokenForWorkspace(
-						requireLinearWorkspaceId(repo),
-					);
-					const issueTracker =
-						this.config.platform === "cli"
-							? (() => {
-									const service = new CLIIssueTrackerService();
-									service.seedDefaultData();
-									return service;
-								})()
-							: new LinearIssueTrackerService(
-									new LinearClient({
-										accessToken: linearToken,
-									}),
-									this.buildOAuthConfig(requireLinearWorkspaceId(repo)),
-								);
-					this.issueTrackers.set(requireLinearWorkspaceId(repo), issueTracker);
-				}
+				// Issue tracker for this workspace is guaranteed to exist here:
+				// updateLinearWorkspaceTokens() runs before addNewRepositories()
+				// and creates trackers for any new workspaces from newConfig.
 
 				// Create activity sink for this repository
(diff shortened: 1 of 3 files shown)


Connoropolous:
@cyrusagent "// New repositories need their workspace's Linear token during
				// initialization, so the token must be available first." why is this the case?

Connoropolous:
@cyrusagent this feels like legacy... shouldn't we consider removing that check?

Connoropolous:
@cyrusagent it seems like we're still constructing the LinearActivitySink there, and doing one per repo... maybe unnecessary, any improvement we can make around that?

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -2629,9 +2618,11 @@ ${taskSection}`;
...
 								);
-					this.issueTrackers.set(
-						requireLinearWorkspaceId(repo),
-						newIssueTracker,
+					const wsId = requireLinearWorkspaceId(repo);
+					this.issueTrackers.set(wsId, newIssueTracker);
+					this.activitySinks.set(
+						wsId,
+						new LinearActivitySink(newIssueTracker, wsId),
 					);
```
Connoropolous: these should go away too.

Connoropolous (changes requested):

Connoropolous:
@cyrusagent is it still true that "// Update Linear workspace tokens BEFORE adding new repositories.
				// New repositories need their workspace's Linear token during
				// initialization, so the token must be available first." 

?

Connoropolous:
@cyrusagent I think you're convoluting what I'm asking for I'm asking is it STILL fundamentally required, or is it just code comments that "say" its required.

Connoropolous (approved):

Connoropolous:
@cyrusagent codex review is right.

can you fix this by not having the deleted repositories code path delete issue trackers and activity sinks

Connoropolous:
@cyrusagent then maybe open a new PR

PR #1167, opened 2026-04-28
fix(security): patch uuid advisory GHSA-w5hq-g745-h8pq (CYPACK-1147)

cyrusagent:
## Summary
- Adds root `pnpm.overrides` entry `"uuid": ">=14.0.0"` to resolve [GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq) (medium, "Missing buffer bounds check in v3/v5/v6 when buf is provided").
- Vulnerable `uuid` versions (8.3.2 / 9.0.1) are pulled in transitively via `@google/gemini-cli-core@0.17.0` (a devDependency, pinned for testing) and `@google-cloud/logging`, `gaxios`, etc. The patched version is a major bump (14.0.0) that no upstream package resolves to yet, so a direct-dep bump cannot reach it — `pnpm.overrides` is the appropriate tool per our dependency policy.
- After the override, `uuid` consolidates to a single resolved version and `pnpm audit` reports zero advisories.

Closes [CYPACK-1147](https://linear.app/ceedar/issue/CYPACK-1147).

## Test plan
- [x] `pnpm install` succeeds and lockfile regenerates
- [x] `pnpm why uuid -r` shows a single resolved version of uuid
- [x] `pnpm audit` reports zero advisories
- [x] `pnpm build` passes
- [x] `pnpm typecheck` passes
- [x] `pnpm test:packages:run` passes (one pre-existing flake unrelated to this change — `DEBUG_CLAUDE_AGENT_SDK` env var leaks from parent shell; passes when env var is unset)


cyrusagent:
> does this dep bump actually break something and do we need to pursue the official major version bump?

Dug into both halves:

### Does it break anything?

No — verified end-to-end on this branch:

- **End-user impact: zero.** `uuid` only appears in our graph via `@google/gemini-cli-core@0.17.0`, which is a **devDependency of `cyrus-gemini-runner`**. Nothing in the published `cyrus-ai` bundle pulls it in. `pnpm why uuid -r` confirms the only path is `cyrus-gemini-runner (devDependencies) → @google/gemini-cli-core → @google-cloud/logging` (+ siblings).
- **CJS consumers still load.** uuid 14 is ESM-only, but Node 22 ships stable `require(ESM)`. I verified `require("eventid")` (which calls `uuid.v4(null, this.b, 8)` — the buffer-write form) and `require("@google-cloud/logging/build/src/utils/context.js")` (which calls `uuid.v4()`) both work with uuid 14 on Node 22.17. CI uses Node 22.
- **Tests pass:** `pnpm build`, `pnpm typecheck`, `pnpm test:packages:run` (incl. the 198 gemini-runner tests) all green.
- **Caveat:** if a contributor runs tests on Node < 22.12 (where `require(ESM)` is not stabilized), the gemini-runner path could fail with `ERR_REQUIRE_ESM`. `apps/f1`'s `engines.node` says `>=18.0.0`. We can either bump that floor or leave it — it doesn't affect shipped users.

### Do we need to pursue the official major bump?

I don't think so, and the underlying risk is also smaller than the advisory implies:

- The advisory ([GHSA-w5hq-g745-h8pq](https://github.com/advisories/GHSA-w5hq-g745-h8pq)) is specifically *"Missing buffer bounds check in **v3/v5/v6** when `buf` is provided."* **None of our consumers call v3/v5/v6.** Both `eventid` and `@google-cloud/logging` only use `uuid.v4()`. So the actual exploitable code path is not exercised in our graph — the override fixes a flagged-but-not-exercised vulnerability and silences Dependabot.
- Pursuing upstream major bumps of `@google/gemini-cli-core`, `@google-cloud/logging`, `eventid`, etc. would be a much wider change for a v4-only consumer set, and most upstreams haven't released a uuid-14-compatible version yet anyway. The override is the precise tool here and per our `CLAUDE.md` policy item #3 should be removed automatically when a future direct-dep bump makes it redundant.

tl;dr — safe to merge, no upstream major bump needed.

cyrusagent:
Superseded by #1181, which addresses both currently-open Dependabot advisories (@anthropic-ai/sdk GHSA-p7fg-763f-g4gf and uuid GHSA-w5hq-g745-h8pq) in a single consolidated PR. Closing in favor of that PR.

PR #1190, opened 2026-05-07
fix(edge-worker): clean up ~/.cyrus/<issue>/ on terminal state to prevent spawn E2BIG

cohenam:
## Summary

- When an issue moves to **Done** or **Cancelled**, also delete the per-issue `~/.cyrus/<issue>/` directory next to the existing worktree cleanup (CYPACK-961 mirror).
- Fixes a class of "stuck session" failures where `spawn E2BIG` (errno -7) blocked **every** new Claude session on long-running deployments.

## Why

Each per-issue subdirectory under `~/.cyrus/` becomes a separate `Read(...)` deny pattern in `buildHomeDirectoryDisallowedTools`, because every `~/.cyrus/<other-issue>/` is a sibling of the single allowed cwd path. After thousands accumulated, the disallowed-tools argv exceeded the kernel `ARG_MAX` limit and **every** `spawn()` of the `claude-agent-sdk` subprocess failed with `errno -7 / E2BIG`. From Linear it looked like sessions were stuck "between subroutines" — issue moved to *In Progress* but never produced a final response.

Concrete numbers from a 4d-uptime production pod:
- 4,668 directories under `~/.cyrus/` (oldest from January, newest today)
- `cqo.disallowedToolsCount` reached **5,031** at session start
- 403 `spawn E2BIG` errors / 0 successful "Session completed" in a 5,000-line log window
- ~100 distinct issues affected before the cleanup was applied

The error was caught and logged but no completion event fired, so Linear stayed in `started` indefinitely.

## Approach

Mirrors the existing `GitService.deleteWorktree(...)` call in `handleIssueStateChangeMessage`. When an issue reaches a terminal state, the per-issue attachments / transient-state directory is reaped on the same trigger as the worktree.

```diff
+import { mkdir, readdir, readFile, rm, writeFile } from "node:fs/promises";
 ...
 this.gitService.deleteWorktree(message.workItemIdentifier);
+
+const issueCyrusDir = join(this.cyrusHome, message.workItemIdentifier);
+try {
+    await rm(issueCyrusDir, { recursive: true, force: true });
+} catch (err) {
+    this.logger.warn(`Failed to remove ${issueCyrusDir} during terminal-state cleanup: ${...}`);
+}
```

This keeps the deny-list size bounded over time. It does not change permission semantics or the structure of `buildHomeDirectoryDisallowedTools` — that remains correct for active issues; the bug was purely state hygiene.

## Operational follow-up (out of scope for this PR)

Existing pods with already-accumulated stale directories need a one-time cleanup to bring `~/.cyrus/` back under `ARG_MAX`. Future restarts after this PR ships will not regrow indefinitely.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm --filter cyrus-edge-worker test:run` — no regressions (13 pre-existing failures in `screenshot-upload-hooks.test.ts` and `runner-selection.test.ts` are present on unchanged HEAD; verified by swapping in the original file and re-running)
- [x] `pnpm build` — clean (also runs as a pre-commit hook)
- [ ] Manual: deploy to a long-running pod, verify a Linear issue moving to Done deletes its `~/.cyrus/<issue>/` directory and that `disallowedToolsCount` does not regrow


Connoropolous:
indeed, I can see a huge issue here. thanks we'll review stat!

PR #1213, opened 2026-05-14
Add Project Update + Project description as agent surfaces (Workstream A)

alicejt:
## What this does

Cyrus today only listens for **Issue**-level Linear events, so the Project surface — the Updates feed and the description body — is invisible to agents. This adds two capabilities:

### A1 — Project Updates as a conversation surface
An `@<agent>` mention inside a Linear **Project Update** spawns a conversational agent session bound to that project, and the agent replies as a follow-up Project Update. It reuses `ChatSessionHandler` (the Slack-chat engine), so the session runs in a plain workspace directory — **no git worktree** — because the surface is discussion, not code. The project *is* the thread: follow-up Updates on the same project resume the same session.

### A2 — Project description as standing context
On a `Project` webhook whose description changed, the description is cached; when an agent later picks up **any** issue under that project, the cached description is injected into its system prompt as long-running context. Caching is opt-in via `CYRUS_PROJECT_CACHE_URL` / `CYRUS_PROJECT_CACHE_TOKEN` env vars — when unset it's a no-op, so this is inert for installs that don't want it.

## Routing — self-filtering, not central routing

`ProjectUpdate` and `Project` webhooks are **workspace-wide**: every installation subscribed to them receives every event (unlike Issue/AgentSession webhooks, which are agent-targeted). So instead of a central router picking an agent, **each instance self-filters** — it resolves its own Linear identity (from `linearAgentId`/`linearAgentName` config, else the `viewer` query at startup) and acts only when it is the agent `@`-mentioned. It never replies to its own Updates (loop prevention).

## Changes by package

- **core** — `ProjectUpdateWebhook` / `ProjectWebhook` types + `isProjectUpdateWebhook` / `isProjectWebhook` / `isProjectDescriptionUpdateWebhook` guards; optional `linearAgentId` / `linearAgentName` config fields.
- **linear-event-transport** — `fetchProject` + `createProjectUpdate` helpers on `LinearIssueTrackerService`.
- **edge-worker** — `handleProjectUpdateWebhook` / `handleProjectWebhook` wired into `handleWebhook`; `LinearProjectChatAdapter` (a `ChatPlatformAdapter` for Project Update conversations); `ProjectDescriptionCache` client; `project-mentions.ts` (mention parsing + the self-filter gate); self-identity resolution at startup; `<project_context>` injection into new-session system prompts.
- Unit tests for the type guards, the mention routing gate, the adapter, and the cache client.

## Subscription / config (operator-facing)

To enable, subscribe the Linear app to the `ProjectUpdate` and `Project` webhook events. `handleWebhook` already ignores unknown event types gracefully, so subscribing before/without this change is harmless. A2's cache is off unless the two env vars are set.

## Testing

- New unit tests pass; full `core` + `edge-worker` suites show no new failures (pre-existing `screenshot-upload-hooks` / `runner-selection` failures on `main` are unrelated).
- `pnpm build` + `pnpm typecheck` clean across the monorepo.
- **Draft** until the end-to-end test plan (direct mention → reply, no-mention quiet, description-as-context, cache invalidation) is exercised against a live workspace.


alicejt:
Validated end-to-end against a live Linear workspace with an isolated test instance:

- **Spike** — confirmed Linear encodes an agent @-mention in a Project Update body as plain `@<name>` text; the self-filter matcher catches it.
- **A1** — `@<agent>` in a Project Update → worktree-free chat session → reply posted as a follow-up Project Update. Loop prevention confirmed (the agent's own reply Update is ignored).
- **A1 no-mention** — a plain Project Update with no mention is received and ignored, no session spawned.
- **A2** — editing a project's description fires a `Project` webhook → description cached; a new issue under that project picks it up and the agent's reply reflects it (verified the agent used the description with **zero tool calls** — it came purely from the injected `<project_context>`). Re-editing the description invalidates the cache; the next issue picks up the new value.

All four test-plan items pass. Marking ready for review.

PR #1221, opened 2026-05-16
refactor: simplify parent session resume callback

### LINKED ISSUE #322: lets make it so that agentSessionManager isn't getting passed as a variable while its its getting defined. It works, but is confusing
https://github.com/ceedaragents/cyrus/blob/a4bc02188aa08a8738ef9f421894a76162821fa9/packages/edge-worker/src/EdgeWorker.ts#L172

lets make it so that agentSessionManager isn't getting passed as a variable while its its getting defined. It works, but is confusing

SSDWGG:
## Summary
- stop passing agentSessionManager into the parent-resume callback while it is being initialized
- have handleResumeParentSession read the shared session manager from the EdgeWorker instance

Fixes #322

## Tests
- pnpm exec biome check packages/edge-worker/src/EdgeWorker.ts
- pnpm --filter cyrus-edge-worker... build
- pnpm --filter cyrus-edge-worker typecheck
- commit hook also ran full monorepo build and typecheck

### DIFF
--- a/packages/edge-worker/src/EdgeWorker.ts
+++ b/packages/edge-worker/src/EdgeWorker.ts
@@ -446,8 +446,6 @@ export class EdgeWorker extends EventEmitter {
 					parentSessionId,
 					prompt,
 					childSessionId,
-					repo,
-					this.agentSessionManager,
 				);
 			},
 		);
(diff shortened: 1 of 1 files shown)


Connoropolous (approved):

Connoropolous:
Thanks for the clean up @SSDWGG !

PR #1224, opened 2026-05-17
fix: harden Linear OAuth refresh against rotation-race / dead-token failures

alicejt:
## Summary

Adds five layered defenses against the refresh-token-death failure class that took out 8 of 13 agents in one Cyrus deployment over a few weeks. Symptoms: an agent's Linear OAuth refresh token ends up invalid (`400 invalid_grant`) and the agent goes silent until manual re-auth — sometimes days later, after the next webhook reveals the dead state.

Root cause is the standard rotation-race + crash-mid-rotation problem class with rotating-refresh-token OAuth servers. The same pattern has been documented by [MCP TypeScript SDK #1760](https://github.com/modelcontextprotocol/typescript-sdk/issues/1760), [OpenAI Codex #10332](https://github.com/openai/codex/issues/10332), and [Anthropic Claude Code #25609](https://github.com/anthropics/claude-code/issues/25609) in the last 12 months. None of those projects had shipped a layered defense at the time this PR was opened — this one does.

## What ships

1. **Per-host advisory file lock** (`proper-lockfile`) on the credentials file wraps the refresh call, so sibling processes sharing a config.json can't race ``/oauth/token``. The Codex and Claude Code issue authors converged on this same approach independently.
2. **Disk re-read after lock acquisition** — if a peer rotated while we were queued, we adopt the persisted token instead of issuing a stale request that would burn it.
3. **Replay sidecars** — the outbound refresh request is persisted to ``<cyrusHome>/state/refresh-pending-<wsId>.json`` *before* being sent. Deleted on success. On startup, ``LinearIssueTrackerService.replayPendingRefreshes`` scans for sidecars and replays any within Linear's [documented 30-minute idempotency window](https://linear.app/developers/oauth-2-0-authentication) — Linear returns the same new tokens for a replay within that window, so this recovers crash-between-send-and-persist deterministically.
4. **Atomic rename** on the workspace-config write (``writeFile`` to a temp file then ``rename``) so a crash mid-persist can't corrupt or truncate the credentials file. POSIX rename is atomic on the same filesystem.
5. **Proactive windowed refresh** — new ``TokenHealthMonitor`` checks every 5 min and triggers refresh when any workspace's access token has <5 min left. Eliminates the thundering-herd concurrent-401 path that's the root cause of the rotation race. Goes through the same single-flight ``pendingRefreshes`` coalescing as reactive refresh.

Plus two related improvements:

- **``invalid_grant`` is now treated as terminal** per [RFC 9700 §4.14](https://datatracker.ietf.org/doc/rfc9700/) — the auth server has revoked the entire refresh-token family; retrying compounds the damage. Now logs an operator-actionable error with the exact re-auth command instead of looping.
- **``TokenHealthMonitor`` viewer-query liveness check** runs alongside the proactive refresh, surfacing dead-token states within 5 min via ERROR-level logs that external journal-based monitors can grep for. Without this, the silent-failure window was up to 7 days (next access-token expiry on an idle agent).

## Why this design (vs alternatives we considered)

- **Centralized broker daemon or shared store** (e.g. Redis, Postgres ``SELECT FOR UPDATE``): overkill for the common single-host fleet case. The file-lock variant is what comparable projects (Codex, Claude Code) independently proposed.
- **Long-lived Personal API Keys** (sidestepping rotation entirely): ruled out — Linear's Agent APIs (``app:assignable``, ``actor=app``, AgentSessionEvent webhooks) are OAuth-app-only. Personal keys can't carry agent identity.
- **Write-ahead marker + dual-token fallback**: tested against the literature and found no precedent. The replay-sidecar pattern + Linear's 30-min grace window cover the same failure modes more cleanly.

## Test plan

- [x] ``pnpm typecheck`` clean across all packages
- [x] ``pnpm build`` clean across all packages
- [x] Existing test suite passes (689/702, the 13 failures are pre-existing in ``EdgeWorker.screenshot-upload-hooks.test.ts`` and unrelated to this PR — verified by running them against ``local-deploy`` without this change)
- [x] Deployed to production VM (13 agents) and confirmed:
  - ``Token health monitor starting`` log line appears on every agent
  - ``invalid_grant`` failures now emit the new terminal-error log with the re-auth command
  - Proactive refresh fires on startup for agents with known token expiry
- [ ] Unit tests for ``LinearIssueTrackerService.replayPendingRefreshes`` and ``TokenHealthMonitor.checkAll`` are TODO for a follow-up PR — the changes here are observably-correct in production but would benefit from regression coverage

## Files changed

- ``packages/linear-event-transport/src/LinearIssueTrackerService.ts`` — refresh path: flock, disk re-read, replay sidecars, ``invalid_grant`` terminal, expiry tracking, ``refreshIfExpiringSoon``, static ``replayPendingRefreshes``
- ``packages/linear-event-transport/package.json`` — adds ``proper-lockfile`` dependency
- ``packages/edge-worker/src/TokenHealthMonitor.ts`` — new module
- ``packages/edge-worker/src/EdgeWorker.ts`` — wires lock target, sidecar dir, disk-recheck callback into OAuth config; starts replay + health monitor in ``start()``, stops monitor in ``stop()``; atomic-rename on ``saveOAuthTokens``
- ``CHANGELOG.md``

## Notes

The 8 agents in the originating deployment whose refresh tokens were already dead before this PR shipped will still need manual re-auth — this hardens against future damage rather than retroactively repairing existing dead state. A follow-up worth considering: an automated re-auth flow that detects ``invalid_grant`` and walks the operator (or the agent itself) through the recovery.

Discussion of the design tradeoffs and prior-art research that led to this approach: https://linear.app/tincture/issue/ENG-44


rairulyle:
Hopefully this get merged asap

PR #1236, opened 2026-05-21
Add slackAllowedTools / githubAllowedTools to EdgeConfig

cyrusagent:
## Summary

Brings the cyrus runtime in line with the per-surface tool + MCP design from CYHOST-967.

### EdgeConfig additions
- `linearAllowedTools` (renamed from `defaultAllowedTools`; legacy field still accepted on parse and migrated forward), `slackAllowedTools`, `githubAllowedTools` — top-level team-level allow-lists, one per surface.
- `linearMcpConfigs`, `slackMcpConfigs`, `githubMcpConfigs` — per-surface arrays of filesystem paths to custom-integration `.mcp.json` files. **Not** a blanket override:
  - **Slack chat sessions** are repo-agnostic — they load exactly the files in `slackMcpConfigs` and never consult `repository.mcpConfigPath`. The prior "first configured repo wins" fallback is gone.
  - **Linear / GitHub / GitLab issue sessions** use the per-platform list only when the routed repo does **not** carry its own `allowedTools` override. If the repo has its own allow-list, the agent uses `repository.mcpConfigPath` instead so the repo's permission rules and its server set always travel as a unit.
  - Native MCP servers (Linear, Cyrus tools, Cyrus docs, and Slack when `SLACK_BOT_TOKEN` is set) are still spun up inline by the runtime — these lists govern custom integrations only.

### ToolPermissionResolver
- Now additive-only — no implicit MCP appending, no `SLACK_BOT_TOKEN`-conditional injection. The explicit per-platform defaults in `cyrus-core` include native MCP prefixes verbatim.
- New `buildGithubAllowedTools()` helper; EdgeWorker's GitHub / GitLab session paths now route through it (subtractive `mcp__slack` filter removed).
- `"readOnly"` preset now resolves to `SLACK_DEFAULT_ALLOWED_TOOLS`.

### `cyrus-core` defaults
- Explicit per-platform defaults exported from `cyrus-core`: `LINEAR_DEFAULT_ALLOWED_TOOLS`, `SLACK_DEFAULT_ALLOWED_TOOLS`, `GITHUB_DEFAULT_ALLOWED_TOOLS`, plus a `getDefaultAllowedTools(platform)` resolver. cyrus-hosted imports the same constants so there is one source of truth.
- `excludeSlackMcp` flag removed from `McpConfigService.buildMcpConfig` — gating Slack MCP injection by the allowed-tools array (single source of truth) is sufficient.

### Docs corrected
A previous revision documented the per-platform MCP config lists as "authoritative — runtime ignores `repository.mcpConfigPath`" with a fallback to legacy first-repo behavior. That framing reflected an earlier design that was corrected and never shipped. JSDoc / inline comments on `ChatRunnerConfigInput`, `IssueRunnerConfigInput`, `ChatSessionHandlerDeps`, `EdgeConfig.{slackMcpConfigs,linearMcpConfigs,githubMcpConfigs}`, and the `EdgeWorker.buildAgentRunnerConfig` call site are now accurate.

Pairs with [cyrusagents/cyrus-hosted#834](https://github.com/cyrusagents/cyrus-hosted/pull/834) which writes these keys via cyrus-hosted's config builder.

Closes [CYHOST-967](https://linear.app/ceedar/issue/CYHOST-967).

## Test plan

- [x] `pnpm typecheck` green
- [x] `pnpm --filter cyrus-core test:run` (129 tests pass — including `json-schema-export`)
- [ ] Deploy a self-host droplet with a config containing `slackAllowedTools: ["Read", "Glob"]` and verify `@mention` chat sessions only see those tools + native MCP prefixes
- [ ] Deploy with a repo carrying `allowedTools` override and confirm Linear sessions on that repo load `repository.mcpConfigPath` rather than `linearMcpConfigs`
- [ ] Deploy with a repo NOT carrying an override and confirm `linearMcpConfigs` is what loads
- [ ] Verify a config without any of the new keys still resolves to prior behavior (legacy `defaultAllowedTools` migrated to `linearAllowedTools`)


Connoropolous:
@cyrusagent I don't see anywhere in this PR that 'githubAllowedTools' helper is actually called? can you confirm / deny this?

Connoropolous (approved):

PR #1273, opened 2026-05-29
feat(slack): follow all replies in threads Cyrus is bound to (CYPACK-1267)

cyrusagent:
## Summary

Makes Cyrus feel "present" in a Slack thread it's been pulled into. After an initial `@mention` binds a thread, **every subsequent reply in that thread is fed into the same agent session** — no need to re-mention Cyrus each time. (CYPACK-1267)

This is the **cyrus (runtime)** half of the change; the **cyrus-hosted** half (webhook forwarding gate + Slack manifest) is in a paired PR.

## What changed

- **`SlackEventTransport`** now accepts `message` events (not just `app_mention`). It applies cheap, conservative structural filters before doing any work:
  - drops the bot's own / other bots' messages (`bot_id`) — prevents reply loops
  - drops subtype events (edits, deletes, joins, `bot_message`, …)
  - drops non-threaded messages (only a threaded reply can belong to a bound thread)
  - **de-dupes** Slack's double-delivery — a message that mentions the bot arrives as *both* an `app_mention` and a `message` event; we collapse identical `(channel, ts)` pairs (10-min TTL) so the thread is prompted once.
- **`ChatPlatformAdapter`** gains an optional `isSessionInitiatingEvent(event)`. **`ChatSessionHandler`** refuses to *start* a new session from a non-initiating event, so a plain reply only ever **continues an already-bound thread** — it never spins one up. (Defaults to "initiating" when the adapter omits the method, so Linear/GitHub behaviour is unchanged.)
- **`SlackChatAdapter`** marks `app_mention` as session-initiating and `message` as follow-up; `SlackMessageTranslator` now maps `message` → user prompt.
- **Slack setup manifest** (`cyrus-setup-slack`) subscribes to `message.channels` / `message.groups` / `message.mpim` / `message.im` (the `*:history` scopes were already present).

## Design notes (SOLID)

- **SRP** — transport owns ingress + filtering + dedup; the adapter owns platform policy (`isSessionInitiatingEvent`); the handler owns session lifecycle.
- **OCP/LSP/ISP** — the new adapter method is optional and defaults to `true`, so existing adapters (and the test adapter) need no change.
- Works for **both ingress paths**: cyrus-hosted proxy forwarding *and* self-host direct Slack delivery (where cyrus is the only gate — hence the dedup + binding gate live here too).

## Testing

- `pnpm test:packages:run` — all green (slack-event-transport 70, edge-worker 673, etc.). New tests cover: threaded message emitted, non-threaded/bot/subtype dropped, app_mention+message dedup, and the session-initiation gate.
- `pnpm typecheck` and `pnpm lint` clean.

## Operational note

**Existing Slack apps must add the four `message.*` bot events in Event Subscriptions** for follow-ups to be delivered.


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
...
+- Cyrus now follows along in Slack threads it's been pulled into: once you @mention it in a thread, every later reply in that thread is fed to the same session automatically — no need to re-mention it each time. To avoid chiming in on every message, Cyrus only replies when a message is a question it can genuinely help with or is addressed to it by name; it stays quiet on side conversation, acknowledgements, and chatter not aimed at it. Replies that do @mention it still work as before, and Cyrus ignores its own messages, edits, and plain channel chatter outside threads it's part of. Thread follow-ups keep working even after Cyrus restarts — it no longer goes silent on plain replies until you re-@mention it. **Existing Slack apps must add the `message.channels`, `message.groups`, `message.mpim`, and `message.im` bot events in their Slack app's Event Subscriptions for this to take effect.** ([CYPACK-1267](https://linear.app/ceedar/issue/CYPACK-1267), [#1273](https://github.com/cyrusagents/cyrus/pull/1273))
```
pauravhp, line 73: We could also briefly mention the new queue approach here, that messages in quick succession will be responded to, together. Users may feel the need to put more effort than needed into their messages (to avoid the rare burst)

PaytonWebber (approved):

PaytonWebber (approved):

PR #1293, opened 2026-06-05
feat(codex): run Codex exclusively via the app-server — streaming input, 0.137 SDK, per-thread sandbox

cyrusagent:
## What & why

Codex's high-level SDK (`codex exec`) has no mid-turn input channel, so when a Linear comment landed while Codex was working, Cyrus had to abort the in-flight turn and restart it (losing progress). This PR moves Codex **entirely** onto the persistent `codex app-server` JSON-RPC protocol, whose `turn/steer` method injects input into the **active** turn — giving Codex the same mid-task responsiveness as the Claude runner.

**Codex now runs exclusively via the app-server.** The old `codex exec` path, the `EdgeWorkerConfig.codexUseAppServer` flag, and the `CODEX_USE_APP_SERVER` env gate have all been removed — there is no longer an opt-in/opt-out; the app-server is the only transport.

> Stacked on `codex/mcp-and-skills-fix` — review/merge that one first.

## How

**SDK bump.** `@openai/codex-sdk` `0.125.0` → `^0.137.0`. The vendored CLI binary moved from `vendor/<triple>/codex/codex` to `vendor/<triple>/bin/codex`, so `resolveCodexBinary` now probes both layouts.

**Refactor.** `CodexRunner` (was a ~1580-line god class) is now a thin orchestrator over focused collaborators behind a single `CodexBackend` strategy:
- `CodexConfigBuilder` (config/model-fallback) · `mcpConfigTranslator` (Cyrus→Codex MCP) · `CodexSkillStager` (skill symlink lifecycle) · `CodexEventMapper` (event→`SDKMessage`).
- `AppServerCodexBackend` emits a backend-neutral `NormalizedCodexEvent` stream, so the mapper is transport-agnostic. `supportsStreamingInput` is now a constant `true`.

**App-server backend.**
- `initialize` → `thread/start` / `thread/resume` → `turn/start` lifecycle; v2 item translation (camelCase → normalized); `turn/steer` (mid-turn input) + `turn/interrupt`.
- Robustness: per-request control-plane timeouts + a turn idle-watchdog (no hung-session-forever); process teardown verified leak-free.
- Parity: MCP servers via `config.mcp_servers`; `outputSchema` on `turn/start`.
- **Process launch & teardown.** Codex is launched via the `@openai/codex` package's public `bin/codex.js` Node shim, resolved from its `package.json` `bin` entry (`@openai/codex` is a direct dependency; `@openai/codex-sdk` is retained only for its TypeScript types — it exposes no app-server transport / `turn/steer`). The launcher owns platform→vendor binary resolution, so we no longer couple to that layout. Because the shim spawns the native binary as a grandchild, the app-server child runs in its own POSIX process group and is torn down with a group-kill, so the native process is reaped immediately (no orphaned processes — re-validated against the soak harnesses).

**Per-thread filesystem sandbox.** When the egress sandbox is enabled, each Codex session runs under a per-thread sandbox that matches the Claude runner's posture: **writes confined to the worktree** (+ allowed dirs), **reads restricted to the worktree + allowed dirs** (home and the rest of the filesystem denied).
- `resolveCodexSandbox` maps Cyrus sandbox intent to a `ResolvedCodexSandbox` union: `workspace-mode` (default — coarse mode, broad reads; via `thread/start.sandbox` + `config.sandbox_workspace_write`) or `profile` (granular, restricted reads).
- Codex `0.128` ([#19449](https://github.com/openai/codex/pull/19449)) removed the inline `SandboxPolicy.readOnlyAccess` read-restriction mechanism and migrated it to **named permission profiles**. The `profile` arm uses that model: `thread/start.permissions: "<id>"` (mutually exclusive with `sandbox`) + `config.permissions.<id>` — a flattened `filesystem` map of path → `read`/`write`/`deny`, keyed by special tokens (`:minimal` = platform defaults, the anti-shell-starvation escape hatch; `:workspace_roots` = worktree; `:tmpdir`; `:slash_tmp`) or absolute paths.

**Wiring.** `addStreamMessage` → `steer`. EdgeWorker's mid-turn dispatch falls back to the resume path if a steer-only backend rejects a message (no dropped comments; no-op for Claude). Slack/chat busy follow-ups are queued and delivered after the turn, and chat threads correctly continue the same session on follow-ups.

## Testing

- **Unit/replay:** 60 codex-runner tests (incl. replay tests from real `codex app-server` transcripts, concurrency-isolation, timeout/watchdog, sandbox serialization), full edge-worker (707) + core (137) suites green; `pnpm typecheck` + `pnpm build` green.
- **Against the real 0.137 binary:** mid-turn steering (obeyed), startup-gap follow-up (incorporated), `thread/resume` across a killed→fresh process (context recalled), and the granular sandbox end-to-end through `CodexRunner` — **allowed write succeeds, outside write blocked, home read blocked, shell runs** (`:minimal` platform defaults).

## Known limitations

- **Chat/Slack sessions are not filesystem-sandboxed.** The per-thread sandbox (and read restriction) applies only to issue sessions with the egress sandbox enabled; `buildChatConfig` never sets `sandboxSettings`, so Codex chat sessions run in the default `workspace-write` mode (writes confined to the chat workspace, but **reads are broad**, including `$HOME`) and Claude chat sessions are unsandboxed. This is pre-existing behavior, not a regression from this PR, and is left as a tracked follow-up (threading `sandboxSettings` into `ChatRunnerConfigInput`/`buildChatConfig` would give chat the same posture).

Multi-repo *workspace* sessions (extra writable roots) are unit-tested (correct profile/params emitted) but not yet validated at runtime — that needs the orchestrator/sub-issue flow, which F1's repo-routing mode doesn't exercise. Tracked as a follow-up before enabling for multi-repo orchestration.

## Notes
- Network permissions are intentionally out of scope (coarse on/off only); Codex's `experimental_network` is Linux/landlock-only and a no-op on macOS.


Connoropolous:
@cyrusagent does this PR have it so that it spins up one and only one app-server, and then N threads?

Connoropolous (approved):

PR #1360, opened 2026-07-02
fix(edge-worker): let repository.model win over the global default when no explicit model is requested

yscivideo:
# fix(edge-worker): let `repository.model` win over the global default when no explicit model is requested

## The bug

The per-repository `model` config field is unreachable. `RunnerSelectionService.determineRunnerSelection()` always resolves `modelOverride` to a value — when no model label or `[model=...]` description tag is present, it fills in the global default (`claudeDefaultModel || defaultModel || "opus"`):

```ts
const resolvedModelOverride =
    modelOverride ||
    defaultModelByRunner[runnerType] ||
    this.getDefaultModelForRunner(runnerType);
```

Then in `RunnerConfigBuilder.buildIssueConfig()`:

```ts
const finalModel =
    modelOverride ||            // ← always set, so…
    input.repository.model ||   // ← …this can never win
    this.runnerSelector.getDefaultModelForRunner(runnerType);
```

The comment on the config says `// Priority order: label override > repository config > global default`, but in practice the repository tier is dead code. The same applies to `repository.fallbackModel` (shadowed by the always-inferred `fallbackModelOverride`).

**Real-world symptom:** a self-hosted user pins `"model": "claude-opus-4-8"` (or any model) on a repository, configures no global default, and every session still launches with the bare alias `"opus"` — which the bundled SDK resolves to whatever it considers latest, silently ignoring the pin. The Linear timeline shows `Using model: claude-opus-4-7` while the config says otherwise.

## The fix

`determineRunnerSelection()` now returns `modelOverride` / `fallbackModelOverride` **only when a model was explicitly requested** (label or description tag). When nothing explicit is present it returns `undefined` for both, letting the existing precedence chain in `buildIssueConfig()` work as documented:

- `model`: explicit label/tag → `repository.model` → global default (unchanged expression, now reachable).
- `fallbackModel`: explicit override → `repository.fallbackModel` (newly reachable, mirroring the fix) → inferred from the final model via `inferFallbackModel()` (extracted to a public method; preserves the existing behavior where a customized default primary steps its fallback down, e.g. default `sonnet` → fallback `haiku`) → runner-default fallback.

The four resume-runner-switch branches in `buildIssueConfig()` (labels changed on an existing session) now clear their overrides instead of re-filling with the switched-to runner's default, so `repository.model` also resolves on continuation turns — safe because the compatibility gate below blocks any foreign-runner repository model from leaking into a resumed session (covered by a dedicated regression test).

## Cross-runner safety

An agent-only selector (label `codex`/`gemini`/`cursor` or `[agent=...]` tag) with no model tag would otherwise fall through to `repository.model`, handing e.g. a Claude model string to a Codex runner. `buildIssueConfig` therefore gates `repository.model` / `repository.fallbackModel` by runner compatibility (via the extracted `inferRunnerFromModel`): a value is applied only when its inferred runner is unknown (custom/proxy model names still pass) or matches the resolved runner; incompatible values are skipped with a debug log and the runner default applies. Cursor is special-cased: it legitimately runs GPT model ids, so codex-inferred models stay compatible on a cursor runner. Covered by regression tests for the codex/gemini label, `[agent=codex]` tag, cursor+GPT, and resume-switch cases.

## Behavior change

One existing expectation changes: when an explicit model label conflicts with the resolved runner type (e.g. labels `["claude", "gpt-5-codex"]`), the conflicting model is reset and `modelOverride` is now `undefined` (previously the global default) — the final model then resolves through `repository.model` / global default downstream, same effective model when no repo pin exists.

## Tests

- Selector: no labels/tags → both overrides `undefined`; explicit labels still resolve as before.
- Builder precedence (wired through the **real** `RunnerSelectionService`): no label + `repository.model` set → repo model wins; explicit label + `repository.model` set → label wins; neither → global default.
- Fallback: customized default primary with no explicit anything → fallback is inferred from the final model (not the hardcoded runner literal), so the fallback can no longer silently equal the primary.

`pnpm test:packages` green (747/747 in edge-worker), `pnpm typecheck` clean, `pnpm lint` clean on touched files.

## Known follow-up (out of scope here)

`inferFallbackModel()` matches bare aliases (`fable`/`opus`/`sonnet`/`haiku`) but not full versioned IDs, so a `repository.model` like `"claude-fable-5"` gets the generic claude fallback (`sonnet`) rather than the one-tier-down `opus`. The fallback is always a valid same-provider model, so this is a quality nit, not a safety issue — a natural follow-up is prefix-matching in `inferFallbackModel`'s claude branch, mirroring what `inferRunnerFromModel` already does.

https://claude.ai/code/session_01AAAaJDxxUWmhtYySCfSQHD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 - Forwarded and shared Slack messages are now included when you @mention Cyrus. Previously, forwarding a message (for example a Sentry alert) into a channel and @mentioning Cyrus passed along only your typed comment — the forwarded message's contents were dropped, so a forward with no comment gave Cyrus nothing to work with. The forwarded content is now part of the prompt. ([#1326](https://github.com/cyrusagents/cyrus/pull/1326))
+- Per-repository `model` and `fallbackModel` configuration is now respected when no model label or description tag is present (previously the global default always won for both), and is only applied when it's compatible with the resolved agent — an agent-only selection (for example a `codex` or `gemini` label with no model tag) no longer hands a foreign-runner model string like a Claude model name to the wrong runner, including on session-continuation runner switches. Cursor sessions still accept GPT model ids directly (for example `gpt-5.4`), since Cursor supports those models natively. ([#1360](https://github.com/cyrusagents/cyrus/pull/1360))
 
 ### Changed
 - Updated `@anthropic-ai/claude-agent-sdk` from `0.3.173` to `0.3.185` and `@anthropic-ai/sdk` from `^0.104.1` to `^0.105.0`, bringing in the latest Claude Code capabilities and bug fixes. ([CYPACK-1346](https://linear.app/ceedar/issue/CYPACK-1346), [#1342](https://github.com/cyrusagents/cyrus/pull/1342))
--- a/packages/edge-worker/src/RunnerConfigBuilder.ts
+++ b/packages/edge-worker/src/RunnerConfigBuilder.ts
@@ -62,6 +62,12 @@ export interface IRunnerSelector {
 	};
 	getDefaultModelForRunner(runnerType: RunnerType): string;
 	getDefaultFallbackModelForRunner(runnerType: RunnerType): string;
+	inferFallbackModel(model: string, runnerType: RunnerType): string | undefined;
+	inferRunnerFromModel(model?: string): RunnerType | undefined;
+	isModelCompatibleWithRunner(
+		model: string | undefined,
+		runnerType: RunnerType,
+	): boolean;
 }
 
 /**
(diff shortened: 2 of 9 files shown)


yscivideo:
Bumping this — it's been quiet for a couple of weeks. Still relevant for us: without it, `repository.model` silently loses to the global default whenever no explicit model is requested. The branch is still cleanly mergeable against main, and the runner-aware compatibility gate is in place for Cursor's `gpt-*` runs. Happy to rebase or adjust the approach if you'd prefer it structured differently.

thilan365:
Hit this independently today on a fresh self-hosted Community-tier setup (macOS, cyrus-ai 0.2.68) — confirms the diagnosis here.

Set `"model": "claude-sonnet-5"` (well, `"sonnet"`) on a single repository entry with no global `claudeDefaultModel` set: every session still launched with `cqo.model: "opus"`. Traced it to exactly what's described — `determineRunnerSelection()` always resolves `modelOverride` via `defaultModelByRunner[runnerType]` when no label/tag is present, so `input.repository.model` never gets consulted in `buildIssueConfig()`. Only setting the *global* `claudeDefaultModel` actually changed behavior.

+1 on getting this merged — the repo-level override is currently silent dead weight for self-hosted users pinning a model per-repo.

PR #1369, opened 2026-07-10
perf(edge-worker): load only essential MCP tools at session start to cut Linear startup latency

TropicalDog17:
## Problem

When a session starts, the agent often spends close to a minute running **tool-search round-trips against Linear's remote MCP server** before it can read or update the issue — a visible stall at the top of every session.

**Root cause:** the official Linear MCP (`mcp.linear.app`) registers ~50 verbose tools. Their combined tool descriptions **on their own** push past the Claude Agent SDK's ~10%-of-context threshold, which silently flips the SDK into MCP tool-search **auto** mode. In that mode *every* MCP tool — including Cyrus's own local ones — is deferred behind an on-demand `ToolSearch`. Because the Linear server is remote, discovering those deferred schemas costs real network round-trips, so turn 1 stalls before the agent can do anything with the issue.

## Fix

Load only the **essential, local** tools into turn-1 context; defer the large remote catalog.

- Mark Cyrus's small, locally-served MCP servers (`cyrus-tools`, `cyrus-docs`) with `alwaysLoad: true`, exempting them from the auto-defer bucket. These are the tools Cyrus's workflow actually depends on (agent-session/feedback management, file uploads, issue relations, docs search) and they're served locally, so there's no round-trip cost and negligible context.
- Leave the remote `linear` server **deferred** (no `alwaysLoad`). Its ~50 tools would otherwise bloat turn-1 context, and the issue essentials the agent needs on turn 1 — title, description, state, priority, URL, and comment threads — are **already injected into the prompt** by the standard session template. The long tail of Linear operations remains reachable on demand via `ToolSearch` when genuinely needed.

The SDK only exposes `alwaysLoad` per-server (not per-tool for remote HTTP servers), so curating the essentials into always-loaded local servers is the mechanism that satisfies both goals: lean turn-1 context **and** no startup tool-search stall for the tools Cyrus relies on.

## Why this is safe

- `mcp__cyrus-tools` / `mcp__cyrus-docs` are already in the default allowed-tool lists — no permission changes.
- No new tools are introduced, so the cyrus-hosted `/settings/tools` catalog needs no update.
- Behavior for the remote Linear catalog is unchanged (still deferred); this only changes *which* servers are eagerly loaded.

## Testing

- Added `packages/edge-worker/test/McpConfigService.alwaysLoad.test.ts` — asserts `cyrus-tools`/`cyrus-docs` are `alwaysLoad`, the remote `linear` server is left deferred, and CLI mode (no Linear client) still always-loads `cyrus-docs`. All 3 pass.
- `McpConfigService` typechecks clean and lints clean.

## Changelog

Added an entry under `## [Unreleased] → ### Changed` describing the faster session startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Changed
-- Faster session startup: Cyrus's own tools (session and feedback management, file uploads, issue relations, and docs search) now load immediately at the start of a session instead of being deferred behind an on-demand tool search. Previously the large Linear tool catalog could push the session over an internal limit that silently deferred *every* tool, adding a noticeable stall — sometimes close to a minute of Linear round-trips — before Cyrus could act on the issue. The essentials Cyrus needs are now available up front, while the full Linear toolset remains available on demand.
+- Faster session startup: Cyrus's own tools (session and feedback management, file uploads, issue relations, and docs search) now load immediately at the start of a session instead of being deferred behind an on-demand tool search. Previously the large Linear tool catalog could push the session over an internal limit that silently deferred *every* tool, adding a noticeable stall — sometimes close to a minute of Linear round-trips — before Cyrus could act on the issue. The essentials Cyrus needs are now available up front, while the full Linear toolset remains available on demand. ([#1369](https://github.com/cyrusagents/cyrus/pull/1369))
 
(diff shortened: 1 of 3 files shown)


TropicalDog17:
Hi can you review for me @Connoropolous

Connoropolous:
have you verified that this has the intended effect?

PR #1374, opened 2026-07-16
fix(slack): only add the receipt reaction when the event will engage

i8ramin:
## Problem

In Slack, Cyrus adds a 👀 (`RECEIPT_REACTION`) to **every** message in any channel it's a member of, and the reaction is never cleared.

`ChatSessionHandler.handleEvent()` calls `this.adapter.acknowledgeReceipt(event)` unconditionally at the top — before the guard that ignores non-initiating events:

- The Slack app subscribes to `message.channels` / `message.groups`, so Cyrus receives an event for every message in channels it's in.
- Each event is acked with 👀 immediately, then dropped as non-initiating.
- The drop `return`s *before* `acknowledgeProcessed()` (the 👀→✅ swap), so the 👀 is never cleared.
- `SlackChatAdapter.isSessionInitiatingEvent` only returns true for `app_mention` / `upstreamGated`, so every ordinary channel message is acked-then-ignored.

Net effect: Cyrus 👀-reacts to unrelated conversations in any channel it's in.

## Fix

Move the acknowledgement below where `threadKey` / `existingSessionId` are computed, and gate it so it only fires when the event will actually engage — a follow-up in a bound thread, or a session-initiating event:

```ts
if (existingSessionId || this.adapter.isSessionInitiatingEvent?.(event) !== false) {
  this.adapter.acknowledgeReceipt(event).catch(...);
}
```

| Event | Before | After |
|---|---|---|
| @mention (initiating) | 👀 | 👀 |
| Follow-up in a bound thread | 👀 | 👀 |
| Ordinary message, unbound thread | 👀 (never cleared) | no reaction |

Platforms whose adapter doesn't implement `isSessionInitiatingEvent` are unaffected — `?.(...) !== false` is true when the method is absent — so GitHub/Linear behavior is unchanged.

## Testing

- `pnpm --filter cyrus-edge-worker exec vitest run test/chat-sessions.test.ts` → **PASS** (33 tests, including the 3 new ones below)
- `pnpm --filter cyrus-edge-worker test:run` (full package suite) → 728 passed / 8 failed. The 8 failures are **pre-existing on `main`** and unrelated to this change — they live in `EdgeWorker.runner-selection.test.ts` (2) and `EdgeWorker.screenshot-upload-hooks.test.ts` (6). With this branch's changes stashed, the same suite reports the identical 8 failures (725 passed); this PR adds exactly +3 passing tests and 0 new failures.
- `pnpm typecheck` → **PASS** (all packages, `tsc --noEmit`)
- `pnpm biome check` on the changed files → **PASS** (no fixes needed)
- Added unit tests (in `packages/edge-worker/test/chat-sessions.test.ts`, `ChatSessionHandler receipt acknowledgement gate`):
  - `acknowledgeReceipt` is **not** called for a non-initiating event in an unbound thread (and the event is ignored — no runner created, no thread bound).
  - `acknowledgeReceipt` **is** called for an initiating event (e.g. an @mention).
  - `acknowledgeReceipt` **is** called for a non-initiating follow-up in an already-bound thread.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
+- Slack: only add the receipt (eyes) reaction when Cyrus will actually engage — a thread it's bound to, or a session-initiating event like an @mention — instead of reacting to every message in any channel it can see and never clearing it. ([#1374](https://github.com/cyrusagents/cyrus/pull/1374))
 - Forwarded and shared Slack messages are now included when you @mention Cyrus. Previously, forwarding a message (for example a Sentry alert) into a channel and @mentioning Cyrus passed along only your typed comment — the forwarded message's contents were dropped, so a forward with no comment gave Cyrus nothing to work with. The forwarded content is now part of the prompt. ([#1326](https://github.com/cyrusagents/cyrus/pull/1326))
 
 ### Changed
--- a/packages/edge-worker/src/ChatSessionHandler.ts
+++ b/packages/edge-worker/src/ChatSessionHandler.ts
@@ -170,18 +170,28 @@ export class ChatSessionHandler<TEvent> {
 				`Processing ${this.adapter.platformName} webhook: ${this.adapter.getEventId(event)}`,
 			);
 
-			// Fire-and-forget acknowledgement (e.g., emoji reaction)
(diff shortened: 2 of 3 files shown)


Connoropolous:
@i8ramin its weird that you're seeing that behaviour. We have cyrus in lots of our channels and do not see it reacting eyes on every message. 

Can you give more evidence of how to reproduce this?

PR #1388, opened 2026-07-28
fix(slack): back-read thread messages posted between mentions

rairulyle:
## What & Why

Re-mentioning Cyrus in an existing Slack thread only passed along the new
message, so anything discussed since the previous mention was lost. The gap is
most visible with thread following disabled, where untagged messages never reach
Cyrus at all — someone answers a question in the thread, Cyrus gets re-tagged,
and it responds as if that exchange never happened.

Sessions now track how far their thread context extends and back-read the
missing window on re-tag.

The governing invariant: **the cursor may only advance past messages the agent
actually received.**

image

## How to Test

Automated:

```bash
pnpm --filter cyrus-edge-worker test:run          # 749 passed
pnpm --filter cyrus-slack-event-transport test:run # 84 passed
pnpm typecheck
```

Coverage lives in `packages/edge-worker/test/chat-sessions.test.ts` and
`packages/slack-event-transport/test/SlackMessageService.test.ts`, and is
behavioural rather than mock-shaped — the cursor-hold cases assert that the same
window is re-requested on the next mention, and the concurrency case races two
mentions through gated reads that resolve out of order, then asserts the third
read starts from the newer position.

Manual, with `slackThreadFollowing` disabled:

1. `@mention` Cyrus in a channel to open a thread and let it reply.
2. Post two or three messages in that thread **without** tagging Cyrus.
3. `@mention` Cyrus again in the same thread.
4. Cyrus's prompt should include the untagged messages under a
   "posted in this thread since you last had context" preamble.
5. Repeat with thread following enabled and confirm the messages are not
   duplicated.

## Checklist

- [x] Self-reviewed the diff
- [x] Tests added / updated (or N/A — explain why)
- [x] No secrets or env values hardcoded
- [x] Docs updated if behavior changed

## Notes for Reviewer

Two bounded residual gaps, called out rather than left implicit:

- **Delivery failure.** The watermark advances once the *read* succeeds, not
  once the runner accepts the prompt. If `start`/`addStreamMessage` fails
  afterwards, that window is skipped. Rolling it back needs care given the
  monotonic guard, so it is deliberately left out of this PR.
- **Very long gaps.** Keeping the newest messages requires walking every page,
  so the walk is capped (10 pages × 200). Beyond ~2000 messages in a single gap
  the read still truncates — a far larger window than the previous 50, but not
  unbounded.

The extra `conversations.replies` call per event is also worth a look: with
thread following *enabled* every thread message already reaches Cyrus, so the
catch-up read is usually redundant there. Gating the read on `app_mention`
events would keep the cost on the path that actually needs it, if you'd prefer
that in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
-- Cyrus now catches up on what was said in a Slack thread between mentions. Previously, @mentioning Cyrus again in a thread it had already replied to passed along only your new message, so anything the team discussed since the last mention was invisible to it — most noticeably with automatic thread listening turned off, where untagged messages are never delivered. Cyrus now back-reads the messages posted since it last had context and reads them before responding. ([#1388](https://github.com/cyrusagents/cyrus/pull/1388))
+- Cyrus now catches up on what was said in a Slack thread between mentions. Previously, @mentioning Cyrus again in a thread it had already replied to passed along only your new message, so anything the team discussed since the last mention was invisible to it — most noticeably with automatic thread listening turned off, where untagged messages are never delivered. Cyrus now back-reads the messages posted since it last had context and reads them before responding. ([#1388](https://github.com/cyrusagents/cyrus/pull/1388)) Thanks @rairulyle for the contribution!
 
 ### Security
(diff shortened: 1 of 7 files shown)


rairulyle:
Requesting for review: @Connoropolous @pauravhp

Thank youuuuu

rairulyle:
Follow-up on this. I've been in this change since last week though I am currently using a patched version. @Connoropolous @pauravhp

Connoropolous (approved):
It's looking good @rairulyle ! Nice improvement

PR #1391, opened 2026-08-03
Expose spawnClaudeCodeProcess on ClaudeRunnerConfig

Pwnion:
The Agent SDK has a `spawnClaudeCodeProcess` option for overriding how the CLI subprocess is created, but `ClaudeRunner` never forwards it — so in practice Claude Code can only run as a local child of the Cyrus process.

**Why I want it.** I'm running Cyrus on Kubernetes. Today that means a single pod hosting every concurrent session: all worktrees on one volume, every `claude` child sharing a HOME, and scaling up means a bigger node rather than more pods. With this option I can pass a function that creates a pod per issue — its own PVC, its own gVisor runtime class — and proxies stdio to it, so each Linear issue gets a real isolation boundary and the cluster does the scheduling. That's a couple hundred lines against `@kubernetes/client-node` on my side and no other changes to Cyrus. The same hook covers remote hosts or a VM pool for anyone who wants sessions off the box running the edge worker.

The change is a straight passthrough: the option is only added to the query when set, so nothing changes if you don't use it. I also added a presence flag to the Sentry option sanitiser, matching how the other opaque fields are handled.

Tests cover the forwarded and absent cases. `tsc --noEmit`, `biome check` and the claude-runner suite (119 tests) pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-_No unreleased changes._
+### Added
+- `ClaudeRunnerConfig` accepts a `spawnClaudeCodeProcess` override, forwarded to the Claude Agent SDK option of the same name. Lets the Claude Code process run somewhere other than a local child process — a container, a Kubernetes pod, a remote host — without forking the runner. ([#1391](https://github.com/cyrusagents/cyrus/pull/1391))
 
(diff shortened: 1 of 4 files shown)


Connoropolous (approved):
this is actually quite eye opening that you're running this.

Connoropolous:
@Pwnion would you be open to hopping on a call sometime soon to chat about this?

PR #1425, opened 2026-08-23
fix(edge-worker): distinguish a cancelled AskUserQuestion from a declined one

Stener1:
## The problem

`AskUserQuestionResult` collapses two opposite outcomes into one shape. A user seeing a question and **declining**, and a question that **never reached a human at all**, both return `answered: false` with a terse reason:

```ts
// cancelPendingQuestion — "Replaced by new question", "Operation was cancelled", …
pendingQuestion.resolve({ answered: false, message: reason });
```

The model receives only this object. So *"replaced by a newer question"* and *"the user said no"* are indistinguishable — and the absence of an answer reads as permission to continue.

The type's own doc has the same conflation: `message` is described as explaining why the request was **"denied"**.

## What it looks like in production

We hit this on a real task. An agent asked whether to proceed, the options were posted to Linear, the question was cancelled unanswered, and the agent **shipped the work anyway** ~19 minutes later.

The output was good, but it had asked for consent and proceeded without it. That's worse than never asking: the Linear thread now reads as though a human approved it. We first tried to fix it with an instruction in the agent's rules telling it never to ask from a subagent and to treat "not answered" as stop. It recurred on the next task, with the rule loaded and in force 1.5 hours earlier — which is what convinced us it needs to be a property of the result, not advice.

## The change

Adds `cancelled?: boolean`, set on every path where no human answered:

| path | `answered` | `cancelled` |
| -- | -- | -- |
| user responds | `true` | — |
| aborted before posting | `false` | `true` |
| cancelled while open | `false` | `true` |
| replaced by a later question | `false` | `true` |
| elicitation failed to post | `false` | `true` |
| no issue tracker | `false` | `true` |
| multi-question rejection | `false` | `true` |

The `message` is now explicit rather than terse, since it's the only part the model actually reads — it states the question was not answered, that this is not approval, and that the agent should stop and report rather than proceed.

This is additive: `cancelled` is optional and nothing existing reads it.

## Deliberately left alone

**Pending questions remain keyed by Linear agent session.** A parent and its subagents still share one slot, so a second question still cancels the first. Notably `createAskUserQuestionCallback` receives the Claude session id and discards it — that's what would disambiguate them:

```ts
return async (input, _sessionId, signal) => {
  // Note: We use linearAgentSessionId (from closure) instead of the passed sessionId
  return this.askUserQuestionHandler.handleAskUserQuestion(input, linearAgentSessionId, …);
};
```

Keying by both would permit concurrent questions per session — a behavioural change deserving its own PR. This fix makes the collision *honest* in the meantime rather than silent. Happy to follow up if you'd want that.

## Tests

The existing suite asserted the old literal strings, so those assertions now check the new contract. Four cases added covering cancelled-vs-answered, replacement, and the multi-question rejection — including a **control** that an answered result does not carry `cancelled`, without which the assertions would pass against a flag set unconditionally.

Verified:
- `tsc --noEmit` clean, `biome check` clean
- edge-worker suite: **753 passing**
- both mutations fail the new tests — removing the flag → 11 failures, reverting the message → 3 — so they aren't vacuous

### DIFF
--- a/packages/edge-worker/src/AskUserQuestionHandler.ts
+++ b/packages/edge-worker/src/AskUserQuestionHandler.ts
@@ -94,6 +94,32 @@ export class AskUserQuestionHandler {
 	 * @param signal - AbortSignal for cancellation
 	 * @returns Promise resolving to the user's answer or denial
 	 */
+	/**
+	 * Build the result for a question that never reached a human.
+	 *
+	 * Every one of these paths previously returned a bare `answered: false` with a
+	 * terse reason, which is the same shape a human declining would produce. The
+	 * model receives only this object, so "replaced by a newer question" and "the
+	 * user said no" were indistinguishable — and the absence of an answer reads as
+	 * permission to continue. Observed in the wild: an agent asked whether to
+	 * proceed, the question was cancelled unanswered, and the agent shipped the
+	 * work anyway, leaving a Linear thread that looks like it was approved.
+	 *
+	 * The message is deliberately explicit rather than terse, because it is the
+	 * only part of this the model actually reads.
+	 */
+	private notAnswered(reason: string): AskUserQuestionResult {
+		return {
+			answered: false,
+			cancelled: true,
+			message:
+				`NOT ANSWERED — ${reason}. No human saw or responded to this question. ` +
+				"This is not approval and must not be treated as one. Do not proceed as " +
+				"though the question was answered; stop and report that you could not " +
+				"obtain a decision.",
+		};
+	}
+
 	async handleAskUserQuestion(
 		input: AskUserQuestionInput,
 		linearAgentSessionId: string,
@@ -105,11 +131,10 @@ export class AskUserQuestionHandler {
 			this.logger.error(
 				`Invalid input: expected exactly 1 question, got ${input.questions?.length ?? 0}`,
 			);
-			return {
-				answered: false,
-				message:
-					"Only one question at a time is supported. Please ask each question separately.",
-			};
+			return this.notAnswered(
+				"only one question at a time is supported, and this call contained " +
+					`${input.questions?.length ?? 0}; ask each question separately`,
+			);
 		}
 
 		const question = input.questions[0]!;
@@ -119,10 +144,9 @@ export class AskUserQuestionHandler {
 
 		// Check if already cancelled
 		if (signal.aborted) {
-			return {
-				answered: false,
-				message: "Operation was cancelled",
-			};
+			return this.notAnswered(
+				"the session was aborted before the question was posted",
+			);
 		}
 
 		// Get issue tracker
@@ -131,10 +155,9 @@ export class AskUserQuestionHandler {
 			this.logger.error(
 				`No issue tracker found for organization ${organizationId}`,
 			);
-			return {
-				answered: false,
-				message: "Issue tracker not available",
-			};
+			return this.notAnswered(
+				"no issue tracker is available for this organization, so the question could not be posted",
+			);
 		}
 
 		// Check for existing pending question for this session
@@ -144,7 +167,7 @@ export class AskUserQuestionHandler {
 			);
 			this.cancelPendingQuestion(
 				linearAgentSessionId,
-				"Replaced by new question",
+				"it was replaced by a newer question on the same session",
 			);
 		}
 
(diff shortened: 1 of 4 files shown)


packages/edge-worker/src/AskUserQuestionHandler.ts
```diff
@@ -94,6 +94,32 @@ export class AskUserQuestionHandler {
...
+	 * terse reason, which is the same shape a human declining would produce. The
+	 * model receives only this object, so "replaced by a newer question" and "the
+	 * user said no" were indistinguishable — and the absence of an answer reads as
+	 * permission to continue. Observed in the wild: an agent asked whether to
+	 * proceed, the question was cancelled unanswered, and the agent shipped the
+	 * work anyway, leaving a Linear thread that looks like it was approved.
+	 *
+	 * The message is deliberately explicit rather than terse, because it is the
+	 * only part of this the model actually reads.
+	 */
```
pauravhp, line 110: This comment is a bit too much for the code. Remove the incident history, move this helper above the existing `handleAskUserQuestion` JSDoc, and keep any description to a single line.

packages/edge-worker/src/AskUserQuestionHandler.ts
```diff
@@ -144,7 +167,7 @@ export class AskUserQuestionHandler {
 			);
 			this.cancelPendingQuestion(
 				linearAgentSessionId,
-				"Replaced by new question",
+				"it was replaced by a newer question on the same session",
```
pauravhp, line 170: Keep the existing `Replaced by new question` wording here. The shared `NO ANSWER` prefix already adds the missing context.

pauravhp (commented):
Hey @Stener1, thanks for catching this! I agree that Cyrus should clearly tell Claude when a question returned no answer instead of making it look like the user denied it.

A few things:

- Remove `cancelled`. `answered` already tells us whether an answer was returned, and `ClaudeRunner` only passes the message to Claude when denying the incomplete tool call, so the new field isn’t used.
- Keep the message factual and short with `NO ANSWER - ${reason}`. Leave out the ‘stop and report’ instruction, and avoid claiming nobody saw the question since Cyrus can only confirm that no response was received. The [Codex review](https://github.com/cyrusagents/cyrus/pull/1425#discussion_r3839448058) has more context.
- Change the `ClaudeRunner` log from ‘User denied AskUserQuestion’ to ‘AskUserQuestion returned no answer,’ while keeping `behavior: "deny"`.
- Add coverage showing that the unanswered message reaches the Claude SDK denial unchanged.
- Add a short changelog entry scoped to making unanswered `AskUserQuestion` results explicit to Claude. Keep it to the user-facing behavior without repeating the implementation or claiming Cyrus prevents the model from continuing.
- Rebase onto current `main`, rerun the build, typecheck, lint and affected tests

I’ve left a few inline comments for the specific wording and comment cleanup too. Tag `@Connoropolous` when its ready and we'll give it a review!

packages/edge-worker/src/AskUserQuestionHandler.ts
```diff
@@ -182,10 +205,9 @@ export class AskUserQuestionHandler {
 		} catch (error) {
 			const errorMessage = (error as Error).message || String(error);
 			this.logger.error(`Failed to post elicitation: ${errorMessage}`);
-			return {
-				answered: false,
-				message: `Failed to present question to user: ${errorMessage}`,
-			};
+			return this.notAnswered(
+				`the elicitation could not be posted to Linear (${errorMessage})`,
```
pauravhp, line 209: Avoid saying the question definitely wasn’t posted, since delivery can be ambiguous after a request failure. Use `Cyrus could not confirm that the question was presented: ${errorMessage}`.

packages/edge-worker/src/EdgeWorker.ts
```diff
@@ -4788,10 +4788,11 @@ ${taskSection}`;
 			this.logger.warn(
 				"Cannot handle AskUserQuestion response without agentActivity",
 			);
-			// Resolve with a denial to unblock the waiting promise
+			// Unblock the waiting promise. Note this is a cancellation, not a
+			// denial: nobody saw the question, so the result must not read as one.
 			this.askUserQuestionHandler.cancelPendingQuestion(
 				agentSessionId,
-				"No agent activity in webhook",
+				"the webhook carried no agent activity, so no response could be read",
```
pauravhp, line 4795: Use `No agent activity in webhook, so no response could be read` and remove the comment above it.

PR #1468, opened 2026-09-09
feat(edge-worker): add Zulip as a chat platform

seongheum-choi:
### Summary

Adds Zulip as a supported chat platform: @mention the bot in a Zulip topic — or DM it — and Cyrus answers in that topic, the way it already does in a Slack thread. A Zulip **topic** is the thread unit, so a session is bound to one `channel + topic` pair.

I built this for a self-hosted deployment where the team's chat is Zulip rather than Slack, and it has been running against a real realm while I worked through the edge cases below. Happy to split it up, drop pieces, or rework the shape if you'd prefer it differently — including the `EdgeWorker` bit noted at the end.

The `ChatPlatformAdapter` interface did almost all of the work here; this is mostly one adapter plus one transport package, mirroring `slack-event-transport`.

### What's in it

- `packages/zulip-event-transport` — `ZulipEventTransport` (registers `POST /zulip-webhook`), `ZulipMessageService` (post, read a topic, add/remove reactions), and the mention-stripping helper.
- `packages/edge-worker/src/ZulipChatAdapter.ts` — `ChatPlatformAdapter<ZulipWebhookEvent>`.
- `zulipMcpConfigs`, threaded through the hot-reload key list and `normalizeConfigPaths`.
- `docs/ZULIP.md`, and a CHANGELOG entry.

Enabled by setting `ZULIP_SITE`, `ZULIP_BOT_EMAIL`, `ZULIP_API_KEY` and `ZULIP_WEBHOOK_TOKEN`. Without all four the transport does not register, so this is inert for everyone who does not use Zulip.

### Three places Zulip's protocol forced a different shape

**It does not sign outgoing webhook requests.** The payload carries the bot's fixed token and that is the only credential available, so verification is a constant-time compare against `ZULIP_WEBHOOK_TOKEN`. Weaker than Slack's HMAC; `docs/ZULIP.md` says so plainly.

**It expects the reply in the HTTP response**, which cannot work for a turn that takes minutes. The transport answers `{"response_not_required": true}` immediately and posts the real reply over REST with the bot's API key — which is also what makes reactions and history reads possible.

**An outgoing webhook bot only hears @mentions and DMs.** There is no equivalent of `slackThreadFollowing`, so I did not add a toggle that could not work. The catch-up read covers the gap instead: on each mention Cyrus reads the newest window of the topic and the cursor selects what the agent has not seen.

That read is always anchored at the newest end rather than at the cursor. `num_before` caps what the server returns rather than describing a window it scans, so the cost is the same for a fifty-message topic and a fifty-thousand-message one — and when a gap outgrows one window, what falls off is its **stalest** messages. Anchoring at the cursor and walking forward truncates from the wrong end and silently drops the most recent messages, which is the opposite of what you want. The agent is told when older messages were omitted.

### Two Zulip behaviours that needed explicit handling

**Message IDs are plain integers, but `ChatSessionHandler` compares thread cursors as strings.** That is correct for Slack, whose timestamps are zero-padded, but `"9" > "10"` — a raw Zulip ID stalls the cursor the moment an ID gains a digit, and the same messages get re-delivered forever. The adapter pads the cursor and unpads it before the ID reaches the API. There is a regression test that fails if the padding is removed.

**Resolving a topic renames it with a `✔ ` prefix rather than setting a flag**, so one continuous conversation arrives under two different names. Keying on the raw name stranded the session and its cursor the moment anyone ticked the checkbox — which I hit in testing, since it is a routine action rather than an edge case. Sessions now key on the name without that marker, while every call that actually addresses Zulip uses the topic's real current name. Renaming a topic outright, or moving messages between topics, still starts a new session; Zulip has no stable topic ID to key on, and that seemed like the right place to stop.

### Deliberately left out

**A message translator.** The Slack transport also emits `InternalMessage`s onto the internal bus, but every chat handler there (`handleSessionStartMessage` and friends) is still a logging placeholder, so a Zulip translator would add core platform-data types and a `source` union member with nothing consuming them. Happy to add it when the bus migration lands, or now if you would rather keep the transports symmetrical.

**A `zulipAllowedTools` key.** `ToolPermissionResolver.buildChatAllowedTools` already governs every chat platform through `slackAllowedTools`; a parallel key would mean threading a platform parameter through two more classes for no behaviour change today.

### Testing

`pnpm test:packages:run`, `pnpm typecheck` and `pnpm lint` are clean. 48 new tests cover token verification, trigger normalization, the narrow and its parameters, cursor selection, self-message exclusion, truncation reporting, thread identity across resolve/unresolve, and the reaction swap.

It has also been exercised end to end against a live Zulip realm: channel topics and DMs, new sessions and resumed ones, the catch-up read picking up messages posted while the bot was not addressed, resolve/unresolve mid-conversation, and the 👀 → ✅ reaction swap. Worth noting for anyone reading the diff: Zulip's name for ✅ is `check`, not `white_check_mark` — Zulip rejects the Slack name with a 400, and `check_mark` is a different emoji.

### One thing I would like your call on

`EdgeWorker.chatSessionHandler` is a single field typed to `SlackWebhookEvent`, and a second chat platform needs somewhere to live. I added a second field plus an `activeChatSessionHandlers` getter, and moved the three aggregate call sites (`isAnyRunnerBusy`, `getAllRunners`, `getAllChatSessions`) onto it, so a third platform touches one getter instead of hunting down call sites. I also pulled the handler dependency object into `buildChatSessionHandlerDeps` so the two registrations do not copy fifty lines between them.

If you would rather have a `Map<ChatPlatformName, ChatSessionHandler<…>>`, or want that refactor to land separately before this, say the word and I will restructure.


Connoropolous (approved):
looks good!!

I will do my own follow up and add it to the readme.

